PR #1, opened 2023-01-20
feat: scaffold

ryanwolhuter:
* add eslint config with prettier and typescript plugin
* add style constants from brand style guide
* add support for svg files as react components
* add and configure styled components
* add global styles
* add and configure storybook
* add and configure fonts


daywiss (approved):
lg

PR #8, opened 2023-01-30
feat: add web worker project

daywiss:

## motivation
looking into performance improvements, web workers may be a really good way to do this

## changes
this adds a proof of concept to build the oracle  client and load it in a web worker as a new project. its bundled with esbuild


ryanwolhuter (approved):
🔥

daywiss:
gonna pause on this, though it is def useful if we go this direction

PR #9, opened 2023-02-03
feat: add basic coms interface for oracle client/server

daywiss:

## motivation
We want to standardize the data layer to be able to plugin multiple sources to the frontend. This allows us to support gql and web3 querying for data.

## changes
This just provides types that we can build to, including standard oracle request type, and client/server interfaces. Frontend could theoretically start building towards this.


ryanwolhuter (approved):
this looks great 👍

PR #12, opened 2023-02-06
feat: add v1 oracle request client and example

daywiss:
## motivation
We need an implementation of our data interface connecting to a real gql endpoint. 

## changes
This adds a hard coded goerli gql service and example of using it in a client in the app. We still need some environment vars, and implementations for v2 and skinny data, and eventually non gql data. 
image

### DIFF
--- a/src/pages/_app.tsx
+++ b/src/pages/_app.tsx
@@ -7,7 +7,6 @@ import {
   white,
 } from "@/constants";
 import { PanelProvider } from "@/contexts";
-import oracle from "@/helpers/oracleSdk";
 import "@/styles/fonts.css";
 import { darkTheme, RainbowKitProvider } from "@rainbow-me/rainbowkit";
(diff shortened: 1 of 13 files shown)


src/pages/_app.tsx
```diff
@@ -7,16 +7,29 @@ import {
...
+    type: OracleType.Optimistic,
+    chainId: 5,
+  },
+]);
+
+// example of using the client. hoook this up in a context / reducer
+Client([gqlService], {
+  requests: (requests) => console.log(requests),
+  errors: (errors) => console.error(errors),
+});
```
daywiss, line 32: this is really the part you will care about @ryanwolhuter

ryanwolhuter: looks good to me

ryanwolhuter (approved):
looks legit 👍

PR #13, opened 2023-02-07
improve: add env parsing to app config

daywiss:

## motivation
oracle client configs could be pretty complicated, we want to make sure the app has a way to import and validate complex configuration. 

## changes
this is a breaking change, but the app will now parse and validate the config, which must be imported as `import {config} from 'constants'`, as opposed to previously just importing the specific value you wanted.


ryanwolhuter (approved):
looks robust

PR #14, opened 2023-02-07
feat: integrate new gql data and endpoints

daywiss:

## motivation
we recently had new subgraphs deployed with new data added to them. we want to integrate this.

## changes
this updates our types to reflect the new data, and adds ability to configure new subgraphs through envs. envs have been added to vercel so this should work immediately.


ryanwolhuter (approved):
looks good 👍  might want to move around and rename some dirs but we can deal with that when everything is in place

PR #15, opened 2023-02-08
feat: add oracle data context

daywiss:

## motivation
we want to bring in oracle data from the graph into react

## changes
this ads a new context/provider which reduces any request updates into a table indexed by request id. this will most likely need to be updated once we know the shape and requirements for the views.


ryanwolhuter (approved):
looks good to me 👍

PR #16, opened 2023-02-09
feat: add gql assertion queries and configs

daywiss:

## motivation
This integrates the new assert gql endpoints to the data interface

## changes
This basically uses the same type as returned from gql, passes it through the client and adds new envs to enable each chain and subgraph


ryanwolhuter (approved):
very cool

PR #17, opened 2023-02-10
feat: error banner

ryanwolhuter:
### Motivation

Add an error banner component with an associated context. I've elected to keep this context separate from the errors that may show in the panel because I expect that we will handle the error states of actions that the user takes in the functions that transform data for the UI. This banner and context is only intended for showing global errors for things like the graph not working or the user being connected to an unsupported network.

**[Story](https://63d101d13121a53332f6218c-jukricodil.chromatic.com/?path=/story/stories-errorbanner--no-messages)**

### Changes

* add error banner component
* add error context

### DIFF
--- a/src/components/ErrorMessage.tsx
+++ b/src/components/ErrorMessage.tsx
@@ -1,45 +1,19 @@
-import { ethersErrorCodes, mobileAndUnder } from "@/constants";
+import { mobileAndUnder } from "@/constants";
 import type { ErrorMessage as ErrorMessageType } from "@/types";
 import NextLink from "next/link";
 import styled from "styled-components";
 
 export function ErrorMessage({ text, link }: ErrorMessageType) {
-  const isEthersError = ethersErrorCodes.some((code) => text.includes(code));
-
-  function ethersErrorLink() {
-    if (!isEthersError) return null;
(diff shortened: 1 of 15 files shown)


src/components/ErrorMessage.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+export function ErrorMessage({ text, link }: ErrorMessageType) {
+  const isEthersError = ethersErrorCodes.some((code) => text.includes(code));
+
+  function ethersErrorLink() {
+    if (!isEthersError) return null;
+
+    const [firstPart, secondPart] = text.split("[");
+
+    const ethersErrorLink = secondPart
```
daywiss, line 14: this kind of ethers specific stuff should be in a utility and processed before this component i would think, so the component could just focus on a single type of data, either that or make 2 error components to handle different data

daywiss (approved):
looks good, just a comment on the design but its not blocking

src/components/ErrorMessage.tsx
ryanwolhuter: yeah good idea

daywiss (approved):

PR #18, opened 2023-02-13
feat: wire up pages

ryanwolhuter:
### Motivation

Bring the components together in the dapp.

### Changes

* add tables to the dapp pages
* add panel to the dapp


daywiss (commented):
we might want to defer this until after my pr goes in. i wire in the data from the oracle v1 queries in a very similar way and there will be conflicts.

src/components/Table/Table.tsx
```diff
@@ -21,15 +21,21 @@ interface Props {
...
     <Wrapper>
       <_Table>
         <Headers page={page} />
         <TBody>
-          {/* When the data is still loading, we show 10 dummy rows with loading skeletons */}
+          {/* When the data is still loading, we show 4 dummy rows with loading skeletons */}
           {isLoading ? (
             <>
-              {Array.from({ length: defaultResultsPerPage }).map((_, i) => (
+              {Array.from({ length: 3 }).map((_, i) => (
```
daywiss, line 38: this probably shoudl be a constant

src/pages/settled.tsx
```diff
@@ -3,9 +3,10 @@ import { useOracleDataContext } from "@/hooks";
 
 export default function Propose() {
   const { settled } = useOracleDataContext();
+
   return (
     <Layout>
-      <Table rows={settled} isLoading={false} page="propose" />
+      <Table rows={settled} isLoading={settled.length === 0} page="settled" />
```
daywiss, line 9: we might need to add specific loading flag, as we arent gauranteed to always ahve values for these pages. what do u think?

daywiss (approved):
awesome thanks

PR #19, opened 2023-02-14
feat: integrate oracle v1 data into pages

daywiss:

## Motivation
We want to integrate the data from gql into our UI pages. 

## Changes
This updates the oracle data context to be a more complete state for the view. This also converts oracle v1 data to the specific view type, but there are many fields missing which we need solutions for. these are documented in this issue https://linear.app/uma/issue/UMA-1182/integrate-oo-request-data-to-ui-view

image


ryanwolhuter (approved):
great work

PR #20, opened 2023-02-15
feat: add assertion data to oracle context

daywiss:
## motivation
We want to include asserter data into the pages

## changes
this adds asserter data in  but in a limited fashion. This adds the reducer and converter required to map data from gql to global oracle state. we currently do not know how to map a majority of the data from asserter graph to what the UI needs. This will need to be addressed in future prs.


ryanwolhuter (approved):
👍

PR #22, opened 2023-02-16
feat: add loading states to query data

daywiss:
## motivation
we want to be able to show loading screens while waiting on data

## changes
we are introducing a convention here, when the data is undefined, it means it is loading, once defined we can say its loaded. we make all oracle queries undefined until we have somethign returned from gql calls and explicitly check this in the view


ryanwolhuter (approved):
nice

PR #23, opened 2023-02-16
feat: filters mobile styles

ryanwolhuter:
### Summary

Add mobile styles for the search bar and filters section

* make filters section responsive
* add filters menu to show on mobile

**[Story](https://63d101d13121a53332f6218c-oekkgpnist.chromatic.com/?path=/story/stories-filters--default)**


src/components/CheckboxList.tsx
```diff
@@ -0,0 +1,57 @@
...
+        <CheckboxesWrapper key={filter}>
+          <Title>{filter}</Title>
+          {Object.entries(items).map(([itemName, { checked, count }]) => (
+            <Checkbox
+              key={itemName}
+              checked={checked}
+              itemName={itemName}
+              count={count}
+              onCheckedChange={({ ...args }) =>
+                onCheckedChange({ ...args, filter: filter as Filter })
```
daywiss, line 29: surprised this needs to be cast

daywiss (approved):
looks great

src/components/CheckboxList.tsx
ryanwolhuter: yeah it's super annoying, happens because Object.entries loses the type of the values it iterates over. https://github.com/microsoft/TypeScript/issues/35101

ryanwolhuter: the alternative is to manually type the result of Object.entries which is even more annoying

PR #24, opened 2023-02-16
feat: add various token queries

daywiss:
## motivation
we need a way to get token data, balances, and approvals

## changes
because we already have this generalized data interface, we can capture new token data through there. this exposes an interface to allow you to "dispatch" queries for balances, approvals and token metadata and collects it in the oracle data context for the view.


ryanwolhuter (approved):
this is freaking awesome well done

PR #25, opened 2023-02-17
feat: oracle query list (table) mobile styles

ryanwolhuter:
### Motivation

On mobile we show a list of cards representing the oracle queries instead of a table. Add a component for this list which conforms to the same api as the table, so that they can be used interchangeably.

**[Story](https://63d101d13121a53332f6218c-uaqddnnrzi.chromatic.com/?path=/story/stories-oraclequeries--propose)**

### Changes

* Add `OracleQueryList` component for mobile
* Add `OracleQueries` component which shows either the table or the list depending on viewport


.eslintrc.json
```diff
@@ -22,6 +22,7 @@
   "rules": {
     "prettier/prettier": "error",
     "comma-dangle": 0,
+    "@typescript-eslint/consistent-type-imports": "error",
```
ryanwolhuter, line 25: This is a very helpful rule. https://typescript-eslint.io/rules/consistent-type-imports/

I've always preferred the clarity of doing `import type` when importing types. It also prevents naming collisions that happen when a type and a component have the same name (like in the `Filter` component which takes data of type `Filter` for example). This rule makes the editor do it automatically. IMO that's how it should work by default

daywiss (approved):
looks great

PR #26, opened 2023-02-17
feat: add provider envs, adjust token api

daywiss:
## motivation
we need providers for each chain to fetch web3 data from contracts starting with tokens

## changes
this adds env parsing, did some basic e2e tests (not shown) to make sure data is coming in from tokens, adjusted token api slightly


ryanwolhuter (approved):
nice

PR #31, opened 2023-02-22
feat: add token approval and interface for actions

daywiss:
## motivation
we need an interface to be able to know what actions we can take, and wrap those actions based on what the current context is

## changes
this adds an action "hook" which dynamically adds things you can do based on the state of the app. in this case its exposing functions to fetch token info, balance and approval, and has a stub for requesting approval from token contract. this may go through further changes to get the api right.

### DIFF
--- a/libs/src/constants.ts
+++ b/libs/src/constants.ts
@@ -1,0 +1,129 @@
+import type { OracleType } from "./oracle2/types";
+import { ethers } from "ethers";
+
+const getAddress = ethers.utils.getAddress;
+type ContractType = OracleType | "Erc20";
+type ContractInfo = {
+  chainId: number;
+  address: string;
+  type: ContractType;
+};
+type ContractInfoList = ContractInfo[];
+export const ContractInfoList: ContractInfoList = [
+  // v1
+  {
+    type: "Optimistic Oracle V1",
+    chainId: 1,
+    address: getAddress("0xc43767f4592df265b4a9f1a398b97ff24f38c6a6"),
+  },
+  {
+    // goerli
+    type: "Optimistic Oracle V1",
+    chainId: 5,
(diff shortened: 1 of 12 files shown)


libs/src/constants.ts
```diff
@@ -0,0 +1,129 @@
...
+
+const getAddress = ethers.utils.getAddress;
+type ContractType = OracleType | "Erc20";
+type ContractInfo = {
+  chainId: number;
+  address: string;
+  type: ContractType;
+};
+type ContractInfoList = ContractInfo[];
+export const ContractInfoList: ContractInfoList = [
```
daywiss, line 12: i think theres a library for this, but hard coding it for now. unfortunately its hard for the graph to return the oracle addresses in the query, so we need to augment the data so that we know what spender to approve when doing token approval

libs/src/oracle2/services/gql/utils.ts
```diff
@@ -53,11 +56,13 @@ export function convertV1(
...
 export function convertAssertion(
   assertion: GqlAssertion,
-  chainId: number
+  chainId: number,
+  oracleAddress: string
 ): Assertion {
   return {
     chainId,
-    oracleType: OracleType.Asserter,
+    oracleType: "Optimistic Asserter",
```
ryanwolhuter, line 64: we should probably change these to OOv3 before we get any more tangled up in this. I need to do the same on the UI type. spoke to chris and we're still going to call the OOv3 output "assertions"

ryanwolhuter (approved):
very nice

libs/src/oracle2/services/gql/utils.ts
daywiss: yes we do need to do this eventually, and that is confusing

PR #33, opened 2023-02-23
fix: rename assert to v3 and other improves

daywiss:
## motivation
asserter has been renamed to oracle v3 at the contract level, we want to propogate this through the ui

## changes
this refactors assert to oracle v3, as well as refactors some of the services to make them a bit more organized. also names all oracles as v1, v2, v3 for consistency.


ryanwolhuter (approved):
nice

PR #34, opened 2023-02-24
feat: filter and search

ryanwolhuter:
### Motivation

We want to be able to search and filter oracle queries. The UI components for doing so are already built.

Add logic for filtering queries by pre-determined keys: chain, project, and oracle type. 
Add logic for fuzzy searching queries by the searchable keys in the `OracleQueryUI` type.

**[Story](https://63d101d13121a53332f6218c-nkurcwtomn.chromatic.com/?path=/story/pages--verify)**

### Changes

* Add `useSearch`, `useFilters`, and `useFilterAndSearch` hooks to be used in the `Filters` component


src/hooks/filters.ts
```diff
@@ -0,0 +1,294 @@
...
+    },
+    filteredQueries: cloneDeep(queries),
+  };
+  const [state, dispatch] = useReducer(reducer, initialState);
+
+  useEffect(() => {
+    dispatch({ type: "make-entries" });
+  }, [queries]);
+
+  function reducer(state: State, action: Action) {
```
daywiss, line 81: not a huge deal, but it looks like this only relies on `queries` in the closure. Theoretically you could move this function out of the hook and use currying like `useReducer(Reducer(queries))` in the hook where `Reducer = queries => reducer`. not a huge deal but in practice its nicer if you can export and test this logic independent of the hook

src/hooks/filters.ts
```diff
@@ -0,0 +1,294 @@
...
+        newState.filters[filterName] = newFilters;
+        newState.checkedFilters = makeCheckedFilters(newState);
+        newState.filteredQueries = makeFilteredQueries(newState);
+
+        return newState;
+      }
+      case "reset": {
+        const newState = cloneDeep(state);
+
+        Object.values(newState.filters).forEach((filter) => {
```
daywiss, line 144: just a nit, but it seems like you could pre-process the filters at the top of the file and then remove all this logic from the reducer, and just clone the processed filters

daywiss (approved):
looks and works great. logic is a bit hard to follow so i mainly just QAd.

we should also consider a future feature: a to clear the "search", clearing filters leaves the search,

src/hooks/filters.ts
ryanwolhuter: great feedback thank you. I've implemented this now, will request review soon

src/hooks/filters.ts
ryanwolhuter: I don't think this would work because the items in each of the filters are generated from the queries. I didn't extract this to a function because here is the only place this logic is used

PR #36, opened 2023-02-24
feat: add full UI interface, page hook

daywiss:
## motivation
We want to create an api for the app that makes it clean to render according to design

## changes
rather than try to implement functionality, this defines the data that we will need for each part of the UI only at the page level. We can fill in this data in subsequent PRs. This also abstracts the page data into a hook which can also be used for eventually filling in the needed data.

### DIFF
--- a/src/types/ui.ts
+++ b/src/types/ui.ts
@@ -6,6 +6,7 @@ import type {
   oracleTypes,
   projects,
 } from "@/constants";
+import type { BigNumber } from "ethers";
 import type { DropdownMenuCheckboxItemProps } from "@radix-ui/react-dropdown-menu";
 import type { ReactNode } from "react";
 export type ActionType =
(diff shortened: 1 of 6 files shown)


src/types/ui.ts
```diff
@@ -50,6 +51,124 @@ export type OracleQueryUI = {
   formattedReward: string | undefined;
   assertion: boolean | undefined;
 };
+export type BigNumberish = string | number | BigNumber;
```
daywiss, line 54: these types describe the interface, basically all the data needed to render elements in the ui according to design, in their most primitive form. I know this doesnt match exactly what you have in the ui components, so this or the other may change slightly, but this gives us a point to start implementing functionality

ryanwolhuter (approved):
👍 👍

PR #38, opened 2023-02-28
feat: add oracle v2 gql data

daywiss:
## motivation
we want to add oracle v2 gql endpoints to app now that its available

## changes
This re-uses the v1 service, afaik the data structures are the same, and adds the env config to enable it


ryanwolhuter (approved):
great job

PR #43, opened 2023-03-07
improve: improve how errors are handled in oracle client

daywiss:
## motivation
services errors are not thrown correctly, we want to know which service is throwing the error in app

## changes
this correclty populates error array with either error or undefined in the same order we initialize services in the client


ryanwolhuter (approved):

ryanwolhuter:
maybe hold off merging until that beast of a thing is in

libs/tsconfig.json
```diff
@@ -0,0 +1,110 @@
+{
```
daywiss, line 1: this needed to be added back because unable to lint without it. shared folder build issues seem to be working with this

PR #44, opened 2023-03-14
feat: wire in actions to table

ryanwolhuter:
### Motivation

We want to be able to compute dynamic values like token info and actions for queries as needed. However, to be sure that this is working in all cases, we need to reliably generate mock data that we can test these features against.

This PR introduces an architecture for mocking all of the graph queries for all of the oo versions and all of the chains. At the moment I have seeded this mock data with values that make sense, but I will soon add more mock data that will allow us to make sure our checks handle incorrect values too.

You can see the mock data in action in the `Pages` stories. The same mock data is presented to each of the pages, and they correctly decide which ones to show and populate data accordingly via the `OracleDataContext`.

**[Story](https://63d101d13121a53332f6218c-trvlslhpzu.chromatic.com/?path=/story/pages-propose--propose)**

### Changes

* Move types and utils that are used by both src and libs into `/shared`
* Add types for raw graph entities and parsed graph entities
* Consolidate fetching and parsing of OOV1, OOV2, and Skinny OO price requests
* Add `query` names to graph queries with the form OOVersionChainName os that we can detect and mock them
* generalize `actions` to `computed` for getting token info and other computed values about queries
* Add a `makeGraphHandlers` function that generates mock graphql handlers for graph queries across all chains and oo versions


.storybook/preview.tsx
```diff
@@ -1,14 +1,16 @@
...
 import React from "react";
 import { WagmiConfig } from "wagmi";
 import { GlobalStyle } from "../src/components/GlobalStyle";
+import { chains, rainbowKitTheme, wagmiClient } from "../src/pages/_app";
 import "../src/styles/fonts.css";
 import "./rainbow.css";
 
-initialize();
+initialize({
+  onUnhandledRequest: "bypass",
```
ryanwolhuter, line 12: this just silences console warnings for api requests that go through without being intercepted by mock service worker.

libs/src/oracle2/services/assertions/gql/factory.ts
```diff
@@ -1,19 +1,22 @@
...
   address: string;
 };
 
 export const Factory =
   (config: Config): ServiceFactory =>
   (handlers: Handlers): Service => {
     async function fetch({ url, chainId, address }: Config) {
-      const requests = await getRequests(url);
-      return requests.map((request) => convert(request, chainId, address));
+      const requests = await getAssertions(url, chainId);
```
ryanwolhuter, line 16: I gave most of these functions more descriptive names

libs/src/oracle2/services/assertions/gql/queries.ts
```diff
@@ -0,0 +1,44 @@
+import { chainsById } from "@shared/constants";
+import type { ChainId, OOV3GraphQuery } from "@shared/types";
+import { makeQueryName } from "@shared/utils";
+import request, { gql } from "graphql-request";
+
+export async function getAssertions(url: string, chainId: ChainId) {
+  const chainName = chainsById[chainId];
+  const queryName = makeQueryName("Optimistic Oracle V3", chainName);
+  const query = gql`
+    query ${queryName} {
```
ryanwolhuter, line 10: giving queries a name like this doesn't change the logic — it just allows us to intercept each of the queries by name with mock service worker

libs/src/oracle2/services/index.ts
```diff
@@ -1,5 +1,4 @@
-export * as oracleV1Gql from "./oracleV1/gql";
-export * as oracleV3Gql from "./oracleV3/gql";
-export * as oracleSkinnyGql from "./oracleSkinny/gql";
```
ryanwolhuter, line 3: the interfaces for skinny and oo-v1 are actually identical. Pablo has a pr coming that rectifies this in the graph schema. for now we can use our mocks to see this working as expected

libs/src/oracle2/services/oracles/factory.ts
```diff
@@ -22,25 +17,10 @@ export const Factory =
...
-        return gqlSkinny.Factory(config)(handlers);
-      }
       if (config.source === "gql" && config.type === "Optimistic Oracle V3") {
-        return gqlV3.Factory(config)(handlers);
+        return gqlAssertions.Factory(config)(handlers);
       }
-      throw new Error(
-        `Unsupported oracle type ${config.type} from ${config.source}`
-      );
+      return gqlPriceRequests.Factory(config)(handlers);
```
ryanwolhuter, line 23: v1, v2 and skinny are all handled by the same logic

libs/src/oracle2/services/priceRequests/gql/queries.ts
```diff
@@ -0,0 +1,63 @@
...
+        proposalLogIndex
+        disputeTimestamp
+        disputeBlockNumber
+        disputeHash
+        disputeLogIndex
+        settlementTimestamp
+        settlementBlockNumber
+        settlementHash
+        settlementLogIndex
+        ${
```
ryanwolhuter, line 49: we only request these fields when we are querying v2 requests

libs/src/oracle2/types.ts
```diff
@@ -1,144 +1,20 @@
...
-  JsonRpcSigner,
+  FallbackProvider,
   JsonRpcProvider,
+  JsonRpcSigner,
+  Provider,
   Web3Provider,
-  FallbackProvider,
 } from "@ethersproject/providers";
 
-export enum RequestState {
```
ryanwolhuter, line 9: these types are moved to @shared

shared/constants/oracle.ts
```diff
@@ -0,0 +1,38 @@
+export const currencies = ["USDC", "ETH", "RY"] as const;
```
ryanwolhuter, line 1: defining these values as constant lists and deriving types from them lets us use these lists to generate exhaustive mocks by iterating over them.

shared/utils/ethers.ts
```diff
@@ -0,0 +1,80 @@
+import { parseFixed } from "@ethersproject/bignumber";
```
ryanwolhuter, line 1: this logic is ported from the voter dapp

shared/types/oracle.ts
```diff
@@ -0,0 +1,63 @@
+import type {
+  chainNames,
+  chainsById,
+  expiryTypes,
+  oracleTypes,
+  projects,
+  requestStates,
+} from "@shared/constants";
+
+export type OracleTypes = typeof oracleTypes;
```
ryanwolhuter, line 10: types where we have a list of values are derived from const arrays

shared/utils/graphql.ts
```diff
@@ -0,0 +1,273 @@
...
+  OOV2GraphEntity,
+  OracleType,
+  ParsedOOV1GraphEntity,
+  ParsedOOV2GraphEntity,
+  PriceRequestGraphEntity,
+  RequestState,
+} from "@shared/types";
+import { BigNumber } from "ethers";
+
+export function handleGraphqlNullableStringOrBytes<
```
ryanwolhuter, line 14: these functions take in their values in the form `{key: value}` so we can easily know what key we are dealing with and show a warning about that specific key if the type passed in is wrong (this is often a source of silent bugs)

shared/utils/graphql.ts
```diff
@@ -0,0 +1,273 @@
...
+      `Wrong type in \`handleGraphqlNullableBoolean\`: Expected boolean or null for ${property}, got ${typeof boolean} instead`
+    );
+  }
+
+  if (boolean === null) return null;
+
+  return boolean;
+}
+
+export function makeQueryName(oracleType: OracleType, chainName: ChainName) {
```
ryanwolhuter, line 66: by using this util function, we can also generate query names in the mocks and have them match exactly

src/constants/env.ts
```diff
@@ -1,10 +1,11 @@
 import assert from "assert";
-import * as ss from "superstruct";
 import { getContractAddress } from "@libs/constants";
+import * as ss from "superstruct";
 
 const Env = ss.object({
   NEXT_PUBLIC_DEFAULT_APY: ss.optional(ss.string()),
   NEXT_PUBLIC_INFURA_ID: ss.string(),
+  NEXT_PUBLIC_DEFAULT_LIVENESS: ss.string(),
```
ryanwolhuter, line 8: added default liveness env var for setting liveness values

shared/constants/oracle.ts
daywiss: i think since this is in shared, we shoudl assume these coudl be real values used in prod, eitehr that or label the file as storybook specific somehow

shared/utils/graphql.ts
```diff
@@ -0,0 +1,280 @@
...
+
+export function handleGraphqlNullableBigInt<
+  EntityKey extends { [key: string]: string | null }
+>(entityKey: EntityKey | null) {
+  if (entityKey === null) return null;
+
+  const [property, bigInt] = Object.entries(entityKey)[0];
+
+  if (!(typeof bigInt === "string" || bigInt === null)) {
+    console.error(
```
daywiss, line 39: why cant we just throw these errors?

src/components/Panel/Panel.tsx
```diff
@@ -37,45 +37,48 @@ const errorBackgroundColor = addOpacityToHsla(red500, 0.05);
...
-    setError,
   } = content ?? {};
 
+  const [error, setError] = useState("");
+  const {
+    fetchCurrencyBalance,
+    fetchCurrencyAllowance,
+    token,
+    fetchCurrencyTokenInfo,
+  } = useComputed(content);
```
daywiss, line 65: this might be able to evnetually be improved by including these functions in the content, so its scoped with the panel data, rather than being globally computed.  nothing you need to change now but just thinking about this

src/hooks/computed.ts
```diff
@@ -0,0 +1,95 @@
...
+  currencyAllowance: BigNumber | null;
+  fetchCurrencyAllowance?: () => void;
+  token: Token | null;
+  fetchCurrencyTokenInfo?: () => void;
+  sendCurrencyApprove?: () => void;
+};
+/**
+ * Takes a query and returns computed values based on the query.
+ */
+export function useComputed(query: OracleQueryUI | undefined) {
```
daywiss, line 21: i think eventually this should be superceded by adding these queries/mutations to each request

daywiss (approved):
thanks for your work on this! looks great, have a few high level comments on strategy, but nothing you need to immediately change.

libs/src/oracle2/services/oracles/factory.ts
ryanwolhuter: understood, will revert

libs/src/oracle2/services/oracles/factory.ts
```diff
@@ -22,25 +17,10 @@ export const Factory =
   (config: Config): ServiceFactory =>
   (handlers: Handlers): Service => {
     const services: Service[] = config.map((config) => {
-      if (config.source === "gql" && config.type === "Optimistic Oracle V1") {
```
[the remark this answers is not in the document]
ryanwolhuter: yeah I completely agree. I was rushing because I was concerned about blocking you and I cut corners that I should not have. I will do better in future

libs/src/oracle2/services/tokens/factory.ts
```diff
@@ -106,6 +103,7 @@ export function Queries(config: Config, emit: Emit) {
         emit({ type: "token", data: { ...config, ...props } });
       })
       .catch((err) => {
+        console.error(err);
```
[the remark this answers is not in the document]
ryanwolhuter: I actually put this in during a test and forgot to remove it, will take it out now

libs/src/oracle2/types.ts
ryanwolhuter: Very much agree. This is something I often struggle with. I'm going to message you later to get some more advice about doing this in the future

shared/constants/oracle.ts
ryanwolhuter: Yeah sorry I think I was unclear. I meant that these are supposed to be the source of truth of the values used in production. then we can use these same values to generate exhaustive mocks of all the possible values we might see in production and make sure that they work in our stories

shared/utils/graphql.ts
ryanwolhuter: two reasons:

throwing the error here fails silently unless you catch it elsewhere (easy to forget)
just throwing the error means you can only see the value and not the key, so you'll just get an `undefined` for example but you won't know which key was undefined. this way you'll know which key has a wrong value

src/types/ui.ts
```diff
@@ -1,56 +1,55 @@
...
-  currency: Currency | undefined;
+  expiryType: ExpiryType | null;
   oracleAddress: string;
-  tokenAddress: string | undefined;
-  formattedBond: string | undefined;
-  formattedReward: string | undefined;
-  assertion: boolean | undefined;
+  tokenAddress: string;
+  formattedBond: string | null;
+  formattedReward: string | null;
```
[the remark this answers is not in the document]
ryanwolhuter: yeah I think so too

PR #46, opened 2023-03-17
feat: contract interactions

ryanwolhuter:
### Motivation

We're using rainbowkit and therefore wagmi for our wallet connection logic, so we might as well leverage that for our contract interaction logic too.

### Changes

* Use wagmi hooks to fetch token info — symbol, decimals, allowance, balance etc.
* Add builder functions that create contract interaction params for each possible action (propose, settle, etc.) for each query
* Add wrapper hook that calls `usePrepareContractWrite`, `useContractWrite`, and `useWaitForTransaction` for each of a query's actions
* Add types for contract interaction params


libs/src/oracle2/services/assertions/gql/factory.ts
```diff
@@ -1,5 +1,6 @@
 import type { ChainId } from "@shared/types";
 import { parseAssertionGraphEntity } from "@shared/utils";
+import type { Address } from "wagmi";
```
ryanwolhuter, line 3: this address type is `0x${string}`. It's what wagmi wants in most cases and its more strict which makes sense to me. I've updated other types to use it as well for compatibility.

reset.d.ts
```diff
@@ -0,0 +1 @@
+import "@total-typescript/ts-reset";
```
ryanwolhuter, line 1: really cool little lib that fixes a lot of ts headaches https://github.com/total-typescript/ts-reset

shared/constants/abi.ts
```diff
@@ -0,0 +1,181 @@
+export const proposePriceAbi = [
```
ryanwolhuter, line 1: these abis are copied from etherscan. I think this is okay because I don't expect these to change, but we can make another plan if that's not gonna work

src/components/Button.tsx
```diff
@@ -54,24 +54,6 @@ export function Button({
   disabled,
   type = "button",
 }: Props) {
-  if (onClick && href) {
```
ryanwolhuter, line 57: this is more a source of headaches than a help. I regret ever adding it.

libs/src/oracle2/services/assertions/gql/factory.ts
```diff
@@ -15,7 +16,7 @@ export const Factory =
     async function fetch({ url, chainId, address }: Config) {
       const requests = await getAssertions(url, chainId);
       return requests.map((request) =>
-        parseAssertionGraphEntity(request, chainId, address)
+        parseAssertionGraphEntity(request, chainId, address as Address)
```
daywiss, line 19: if we are gonna enforce this type, there should probably be some kind of utility "isAddress" so we dont need to cast

daywiss (approved):
looks good, i think ill have a better idea after doing some work on it

PR #50, opened 2023-03-21
feat: wire in transaction notifications

daywiss:
## motivation
we need to hook up contract interactions to notification system

## changes
this adds a helper which integrates into the wagmi contract writer ape, hooks into onMutate, onSuccess and onError.

@ryanwolhuter this is blocked until i can figure out how to get the notifications to render

### DIFF
--- a/src/components/Notifications.tsx
+++ b/src/components/Notifications.tsx
@@ -40,7 +40,7 @@ export function Notifications() {
 const AnimatedNotification = animated(Notification);
 
 const Wrapper = styled.div`
-  z-index: 1;
+  z-index: 2;
   position: fixed;
   top: 0;
   left: 0;


ryanwolhuter:
@daywiss it's not rendering because you haven't added the `Notifications` react component to the app 😂 there's nothing to render in response to these events. 

I'll take over this PR and add it

src/hooks/actions.tsx
```diff
@@ -292,6 +217,67 @@ export function useActions(
     ),
   ];
 
+  useEffect(() => {
```
ryanwolhuter, line 220: this is pretty much exactly how the voter dapp does it, only difference is now all the handlers live here instead of in individual contract call handler functions

PR #51, opened 2023-03-21
feat: add url query links in navbar

daywiss:
## motivation
we want to be able to link to specific queries in app

## changes
this is a wip, basically changes url to past in tx hash and log index when table is clicked. clears it out when panel is hidden. then we can use this data to filter over all requests.


src/components/Panel/PanelBase.tsx
```diff
@@ -12,18 +10,8 @@ interface Props {
...
   const isClient = useIsClient();
 
-  useEffect(() => {
-    router.events.on("routeChangeStart", closePanel);
-
-    return () => {
-      router.events.off("routeChangeStart", closePanel);
-    };
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
```
daywiss, line 25: this was causing infinite loop, things seem to work without though may need closer inspection

shared/types/graphql.ts
```diff
@@ -405,20 +405,20 @@ export type ParsedOOV1GraphEntity = {
...
-  disputeBlockNumber: BigNumber | null;
+  disputeBlockNumber: string | null;
   disputeHash: string | null;
-  disputeLogIndex: BigNumber | null;
+  disputeLogIndex: string | null;
   settlementTimestamp: string | null;
-  settlementBlockNumber: BigNumber | null;
-  settlementLogIndex: BigNumber | null;
+  settlementBlockNumber: string | null;
+  settlementLogIndex: string | null;
```
daywiss, line 421: these all need to be strings so we can easily compare them

daywiss:
@ryanwolhuter im having wierd errors in dev, maybe you have an idea of whats going on 
image

ryanwolhuter:
@daywiss I don't think we should try and rush in this feature before they do their QA. They are going to be more than busy enough with the contract interactions.

Regarding that error — hydration problems usually happen when the app is trying to do something server side that it can only do client side. I'd guess that the cause here is something to do with the [`Portal`](https://www.radix-ui.com/docs/primitives/utilities/portal) we use for the panel. If you take a look at that implementation, you'll see that it only makes sense to do something like that in the browser. When you run into something like this, 99% of the time all you need is to use the [`useIsClient`](https://usehooks-ts.com/react-hook/use-is-client) hook and do `if (!isClient) return null` which you've probably seen peppered around the app. I didn't get that error when I pulled your branch or on the preview deployment though, which is odd.

For this feature to work I think it still needs two things:

* When visiting a url like this, we should navigate to the page that query is on, so that when the user closes the panel they are able to click the query and open it again.
* We need a way to click on a given query and do a "copy url" type action so that people can have a url to share.

ryanwolhuter (approved):

daywiss:
@ryanwolhuter thanks. i've talked to sean and brought up potential issues, hes good with merging this as is and testing it for further improvements. see issues in [this card](https://linear.app/uma/issue/UMA-1302/have-linkable-urls-for-different-search-filtering-results)

PR #59, opened 2023-03-30
fix: wrong entries and double counting in filters

ryanwolhuter:
### Motivation

There are some bugs in the filtering logic now that we're sharing them across pages. These come about mostly because we are storing the filtering data in a hook, which is being invoked multiple times on different pages. This is solved by lifting this state up into a context and providing that state in the hook instead. The hook can then be invoked wherever.

Another difficulty was passing the page we're on to the filters and making sure that the filters are operating on the right queries. This is solved by removing lifting this state up in a context as well. Now pages explicitly set the "page" value when they are navigated to, and the dependency on the url is removed (helpful for future tasks too).

Finally, the counting logic in the filters was wrong and lead to double counting of entries on renders. This has been fixed with a solution that is also more efficient.

### Changes

* Add `FilterAndSearchContext` for holding filter and search state
* Add `PageContext` for holding which page we are on
* Add hook to get queries for the current page
* Fix entry counting bug in filters

### DIFF
--- a/src/components/Layout.tsx
+++ b/src/components/Layout.tsx
@@ -6,16 +6,15 @@ import {
   Panel,
 } from "@/components";
 import { siteDescription, siteTitle } from "@/constants";
-import { capitalizeFirstLetter, determinePage } from "@/helpers";
+import { capitalizeFirstLetter } from "@/helpers";
+import { useHandleQueryInUrl, usePageContext } from "@/hooks";
 import Head from "next/head";
-import { useRouter } from "next/router";
 import type { ReactNode } from "react";
 import styled from "styled-components";
 
 export function Layout({ children }: { children: ReactNode }) {
-  const router = useRouter();
-  const pathname = router?.pathname;
-  const page = determinePage(pathname);
+  const { page } = usePageContext();
+  useHandleQueryInUrl();
 
   return (
     <>
@@ -28,7 +27,7 @@ export function Layout({ children }: { children: ReactNode }) {
       <Main>
         <ErrorBanner />
         <Header page={page} />
-        <Filters page={page} />
+        <Filters />
         {children}
         <Panel />
         <Notifications />
--- a/src/contexts/FilterAndSearchContext.tsx
+++ b/src/contexts/FilterAndSearchContext.tsx
@@ -1,13 +1,11 @@
 import { emptyCheckedFilters, emptyFilters } from "@/constants";
-import { determinePage } from "@/helpers";
-import { useFilterAndSearch, useOracleDataContext } from "@/hooks";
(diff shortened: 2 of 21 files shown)


src/components/Layout.tsx
```diff
@@ -6,16 +6,15 @@ import {
...
-import { useRouter } from "next/router";
 import type { ReactNode } from "react";
 import styled from "styled-components";
 
 export function Layout({ children }: { children: ReactNode }) {
-  const router = useRouter();
-  const pathname = router?.pathname;
-  const page = determinePage(pathname);
+  const { page } = usePageContext();
+  useHandleQueryInUrl();
```
ryanwolhuter, line 17: handling queries from the url is invoked here in the shared layout component which is persisted between page changes, so its only invoked once

src/contexts/FilterAndSearchContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+  },
+  onCheckedChange: () => null,
+  reset: () => null,
+};
+
+export const FilterAndSearchContext = createContext(
+  defaultFilterAndSearchContextState
+);
+
+export function FilterAndSearchProvider({ children }: { children: ReactNode }) {
```
ryanwolhuter, line 40: this is the same filter and search logic, just lifted up to a context so that we can share it more easily

src/hooks/filters.ts
```diff
@@ -122,44 +118,42 @@ export function useFilters(queries: OracleQueryUI[]) {
...
+}
+
 function filtersReducer(queries: OracleQueryUI[]) {
   // use currying to capture queries
   return function reducer(state: State, action: Action) {
     switch (action.type) {
       case "make-entries": {
         const newState = cloneDeep(state);
 
+        newState.filters.project.All.count = queries.length;
```
ryanwolhuter, line 136: this logic was causing double counting. This new implementation is much simpler and fixes the bug. We don't need to iteratively count items, we can just get the length once and be done with it

src/hooks/handleQueryInUrl.ts
```diff
@@ -0,0 +1,28 @@
...
+  usePageContext,
+  usePanelContext,
+  useQueriesCurrentForPage,
+} from "@/hooks";
+import type { OracleQueryUI } from "@/types";
+import filter from "lodash/filter";
+import { useRouter } from "next/router";
+import { useEffect } from "react";
+
+export function useHandleQueryInUrl() {
```
ryanwolhuter, line 11: logic is the same, just renamed to prevent confusion with other page-related hooks

daywiss (approved):
looks good, working for me

PR #61, opened 2023-03-30
improve: refac organization to include web3 logic

daywiss:
## motivation
want to prepare to add web3 logic, currently the folder structure makes it hard to accomodate this

## changes
the current folder structure is informed by the data shape from graphql, its difficult to say if this same assumption will apply to web3 services. this is refactored to allow for separte logic between all versions of oracle, rather than grouping by request type for flexbility. we can always refactor and remerge these concepts in the future


ryanwolhuter (approved):
I'm happy with this logic, but the directory names are a bit confusing because we've got `oracle1` inside a dir called `oracle2` etc. Can we rather use names like `oov1` or similar inside of the `oracle2` dir to avoid the ambiguity?

daywiss:
> I'm happy with this logic, but the directory names are a bit confusing because we've got oracle1 inside a dir called oracle2 etc. Can we rather use names like oov1 or similar inside of the oracle2 dir to avoid the ambiguity?

yea good point, in my mind the higher level directories are kind of temporarily named, but let me propose a fix here

PR #62, opened 2023-03-30
improve: refac request/assertion types to have more partial vals

daywiss:
## motivation
the shared types for requests and assertions does not work for web3 data, because these queries cannot gaurantee all the properties required.

## changes
this makes more properties optional for requests and assertions so that we can issue partial updates from sources other than the graph.  web3 data may only be dealing with a subset of events, and build an incomplete picture of an oracle request, or may be dealing with calls to contract which may only return partial data, so we need to ensure these types respect that.

### DIFF
--- a/shared/types/oracle.ts
+++ b/shared/types/oracle.ts
@@ -6,6 +6,35 @@ import type {
   projects,
   requestStates,
 } from "@shared/constants";
+import type { Address } from "wagmi";
+import {ParsedOOV1GraphEntity,ParsedOOV2GraphEntity,ParsedOOV3GraphEntity} from "./graphql";
+
+export type Request = {
+  chainId: ChainId;
+  oracleAddress: Address;
+  oracleType: OracleType;
+  id: string;
+  identifier: string;
+  ancillaryData: string;
+  time: string;
+  requester: Address;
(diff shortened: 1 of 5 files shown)


shared/types/oracle.ts
```diff
@@ -6,6 +6,35 @@ import type {
...
+  chainId: ChainId;
+  oracleAddress: Address;
+  oracleType: OracleType;
+  id:string,
+  identifier: string;
+  asserter: string;
+  claim: string;
+  assertionTimestamp: string;
+  assertionId: string;
+} & Partial<ParsedOOV3GraphEntity>;
```
daywiss, line 33: these new types are basically the same but ive tried to cureate data that we know will exist even in the least informed calls for data.  these assumptions may break in the future, but at least we can easily pull out required props.

shared/utils/graphql.ts
```diff
@@ -266,7 +266,7 @@ export function parseAssertionGraphEntity(
   } = assertion;
   return {
     chainId,
-    oracleType: "Optimistic Oracle V3",
+    oracleType: "Optimistic Oracle V3" as const,
```
daywiss, line 269: everythign below here is basically the fallout from making more props undefined, we have to check more values exist now

ryanwolhuter (approved):
nice work, makes sense

PR #63, opened 2023-03-31
feat: project specific metadata

ryanwolhuter:
### Motivation

For our partner projects, we'd like to show some customized details like a project logo, title, description etc. Each of these has its own pretty fiddly solution, which will likely need to be strengthened in the future. None of this stuff is critical though, just cosmetic, so its not super serious if things are occasionally a bit off.

### Changes

* Add parsing logic that identifies some projects and enriches the data passed to the ui with context aware data

### Notes

I've used logos dug up from other projects for now because I need custom ones from Jesper and he's on leave. I'll get them from him when he's back. So don't be alarmed if the logos look a little janky

Right now we don't have a clear way of getting a title and description for the assertions that have been made on goerli. We'll need to ask partners to follow some parsable pattern for these.


package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "next dev",
     "build": "next build",
-    "prebuild": "yarn update-approved-identifiers",
+    "predev": "yarn update-approved-identifiers",
```
ryanwolhuter, line 8: running this before dev instead of build because its extremely fast and this way we'll never have out of date data

daywiss (approved):
looks great, working for me

PR #64, opened 2023-03-31
feat: update oracle data from transaction receipts

daywiss:
## motivation
we want to be able to update our oo query data when users change the state through transactions

## changes
This only adds oo v1 realtime updates based on transaction reciepts. It builds this into the service architecture so that updates are issued through the same logic defined in the oracle data context.  it hooks into the actions hook, and waits for txs to complete before passing these receipts into the service. 

## testing
this is difficult to test currently, and probably is not working correctly just yet. this will be followed up with prs to help refactor and add the rest of the oracles.

### DIFF
--- a/libs/src/oracle-sdk-v2/services/oraclev1/ethers/factory.ts
+++ b/libs/src/oracle-sdk-v2/services/oraclev1/ethers/factory.ts
@@ -1,0 +1,152 @@
+import Events from "events";
+import { ethers } from "ethers";
+import { assertAddress } from "@shared/utils";
+import type {
+  Request as SharedRequest,
+  OracleType,
+  ChainId,
+  RequestState as SharedRequestState,
+} from "@shared/types";
+import type { Address } from "wagmi";
+import type {
+  Handlers,
+  Service,
+  ServiceFactory,
+} from "@libs/oracle-sdk-v2/types";
+import type { TransactionReceipt } from "@libs/types";
+import type { Request } from "@libs/clients/optimisticOracle";
+import { RequestState, requestId } from "@libs/clients/optimisticOracle";
+import { OptimisticOracle } from "@libs/oracle-sdk-v1/services/optimisticOracle";
+
+export type Config = {
+  chainId: ChainId;
+  url: string;
+  address: string;
+};
+
+function convertToSharedState(state: RequestState): SharedRequestState {
+  if (state === RequestState.Invalid) return "Invalid";
+  if (state === RequestState.Requested) return "Requested";
+  if (state === RequestState.Proposed) return "Proposed";
+  if (state === RequestState.Expired) return "Expired";
+  if (state === RequestState.Disputed) return "Disputed";
+  if (state === RequestState.Resolved) return "Resolved";
+  return "Settled";
+}
+
+const ConvertToSharedRequest =
+  (chainId: ChainId, oracleAddress: Address, oracleType: OracleType) =>
+  (request: Request): SharedRequest => {
+    const {
+      requester,
+      identifier,
+      timestamp,
+      ancillaryData,
+      currency,
+      reward,
+      finalFee,
+      proposer,
+      proposedPrice,
+      expirationTime,
+      disputer,
+      price,
+      settleTx,
+      requestTx,
+      proposeTx,
+      disputeTx,
+      requestBlockNumber,
+      proposeBlockNumber,
+      disputeBlockNumber,
+      settleBlockNumber,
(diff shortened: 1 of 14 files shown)


src/constants/env.ts
```diff
@@ -486,70 +495,108 @@ function parseEnv(env: Env): Config {
...
+      url: env.NEXT_PUBLIC_PROVIDER_V1_80001,
       chainId: 80001,
+      address: getContractAddress({
+        chainId: 80001,
+        type: "Optimistic Oracle V1",
+      }),
     });
   }
-  // verify we have providers for all enable subgraph chains. If not checked we will have run time errors if providers are missing.
-  subgraphs.map((subgraph) => {
```
daywiss, line 544: this is no longer needed since we rely on wagmi to pull meta data and not the oracle services. each provider can now be defined to enable or disable this real time update system per chain per oracle.

src/hooks/actions.tsx
```diff
@@ -209,8 +210,15 @@ export function useProposeAction({
       pending: <>Proposing price</>,
       success: <>Proposed price</>,
       error: <>Failed to propose price</>,
-    }).catch(console.error);
-  }, [proposePriceTransaction, chainId]);
+    })
+      .then((receipt) => {
+        if (receipt && query)
+          oracleEthersApis?.[query.oracleType]?.[
```
daywiss, line 216: this initiates the update by searching for the correct service api for the type and chain

libs/src/oracle-sdk-v2/services/oraclev1/ethers/factory.ts
```diff
@@ -0,0 +1,152 @@
...
+    config.chainId,
+    assertAddress(config.address),
+    "Optimistic Oracle V1"
+  );
+  const provider = new ethers.providers.JsonRpcProvider(config.url);
+  const oo = new OptimisticOracle(provider, config.address, config.chainId);
+  const events = new Events();
+  function updateFromTransactionReceipt(receipt: TransactionReceipt) {
+    oo.updateFromTransactionReceipt(receipt);
+    const requests: Request[] = oo.listRequests();
```
daywiss, line 136: this re-uses logic from the sdk-v1, where the oracle client already had ability to decode receipts and update request data. we update the class and fetch all known requests

libs/src/oracle-sdk-v2/services/oraclev1/ethers/factory.ts
```diff
@@ -0,0 +1,152 @@
...
+  if (state === RequestState.Invalid) return "Invalid";
+  if (state === RequestState.Requested) return "Requested";
+  if (state === RequestState.Proposed) return "Proposed";
+  if (state === RequestState.Expired) return "Expired";
+  if (state === RequestState.Disputed) return "Disputed";
+  if (state === RequestState.Resolved) return "Resolved";
+  return "Settled";
+}
+
+const ConvertToSharedRequest =
```
daywiss, line 37: tedious to write, but this ensures the data we compile through events matches the shared type expected in oracle data context

libs/src/utils.ts
```diff
@@ -142,3 +143,51 @@ export async function estimateBlocksElapsed(
...
+export function insertOrderedAscending<T>(
+  array: T[],
+  element: T,
+  orderBy: (element: T) => string | number
+): T[] {
+  const index = sortedLastIndexBy(array, element, orderBy);
+  array.splice(index, 0, element);
+  return array;
+}
+export function isUnique<T>(
```
ryanwolhuter, line 183: doesn't lodash already have this?

libs/src/utils.ts
```diff
@@ -142,3 +143,51 @@ export async function estimateBlocksElapsed(
...
+ * @param {T} element
+ * @param {Function} orderBy
+ */
+export function insertOrderedAscending<T>(
+  array: T[],
+  element: T,
+  orderBy: (element: T) => string | number
+): T[] {
+  const index = sortedLastIndexBy(array, element, orderBy);
+  array.splice(index, 0, element);
```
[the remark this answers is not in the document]
daywiss: good call, at the moment this is used only internally to the oracle service class.  anything that gets passed through to react is returned as a new array

libs/src/utils.ts
daywiss: the "uniq" calls in lodash are about creating lists with unique elements, not sure theres one specifically for this behavior where we want to identify that a value exists in a list using a special identifying function

PR #66, opened 2023-04-03
feat: views for queries that are already proposed or settled

ryanwolhuter:


src/constants/actions.ts
```diff
@@ -0,0 +1,35 @@
+export const propose = "Propose";
```
ryanwolhuter, line 1: I added these definitions so we can compare the title strings without worrying about them being wrong after being changed in the future

src/hooks/handleQueryInUrl.ts
```diff
@@ -1,27 +1,37 @@
...
     const hasQueryUrl = Object.keys(router.query).length > 0;
-    const queryResults: OracleQueryUI[] = filter<OracleQueryUI>(
-      queries,
-      router.query
-    );
-    if (hasQueryUrl && queryResults.length === 1 && !panelOpen) {
-      openPanel(queryResults[0], page, false);
+    if (!hasQueryUrl) return;
+
+    const query = find<OracleQueryUI>(queries, router.query);
```
daywiss, line 18: the reason filter was used here is because its possible to search on parameters which might match multiple queries. if thats the case, im not sure we should pick one for the user. `find` will just return the first one found. not a huge deal and we can debate for future pr

src/contexts/PanelContext.tsx
```diff
@@ -30,49 +27,31 @@ export const PanelContext = createContext<PanelContextState>(
...
-        : isAssertion
-        ? {
-            assertionHash,
-            assertionLogIndex,
-          }
-        : {};
+      const query = makeUrlParamsForQuery(content);
 
-      router.push({ query }).catch(console.error);
+      await router.push({ query });
```
daywiss, line 40: this isnt caught anywhere, dont know if it will ever throw tho

daywiss (approved):
minor things but i think we can get this in

PR #67, opened 2023-04-04
feat: add v3 receipt parsing

daywiss:
## motivation
we want to have state updates when user submits txs that change oracle v3 queries

## changes
this does a lot of similar things to the v1 receipt update logic, but we needed to specifically add v3 event to state parsers since we didnt have those from the original sdk. we also have to loosen the UI types because there is some data that is not gauranteed when parsing from events.

### DIFF
--- a/src/types/ui.ts
+++ b/src/types/ui.ts
@@ -60,22 +60,23 @@ export type OracleQueryUI = {
   project: Project;
   title: ReactNode;
   description: ReactNode;
-  identifier: string;
   expiryType: ExpiryType | null;
   oracleAddress: Address;
+  // not always available with assertions
+  identifier?: string;
   // price requests query text is the ancillary data
   // for assertions it is the `claim` field
-  queryTextHex: string;
-  queryText: string;
+  queryTextHex?: string;
+  queryText?: string;
   // for price requests the value text is null until a price is proposed. Then it is the proposed price. After a price is settled it is the settled price.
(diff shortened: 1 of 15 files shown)


src/types/ui.ts
```diff
@@ -60,22 +60,23 @@ export type OracleQueryUI = {
   project: Project;
   title: ReactNode;
   description: ReactNode;
-  identifier: string;
   expiryType: ExpiryType | null;
   oracleAddress: Address;
+  // not always available with assertions
+  identifier?: string;
```
ryanwolhuter, line 66: @md0x I think the subgraph schema for v3 is wrong here. The schema says the `identifier` is non-nullable, but for v3 it can be empty

ryanwolhuter (approved):
excellent work — just make sure all the env vars are correct on both the mainnet and goerli vercel envs or all the builds will fail

src/types/ui.ts
daywiss: @ryanwolhuter its possible this will always be available from the graph, but when generating state from an event, its not gauranteed to exist. we may only have a settle or dispute event, which will not include much of the info. the only thing gauranteed from events is the `assertionId`.

ryanwolhuter: Yeah I say this because I can see assertions with null for the identifier coming from the graph, on goerli at least.

PR #69, opened 2023-04-04
feat: add tooltips component

ryanwolhuter:
Add a `Tooltip` component that can be used to show additional information on hover / focus.

It can wrap arbitrary content and shows the tooltip.

The content that goes into the tooltip can be any react code.

[Stories](https://63d101d13121a53332f6218c-jjwgrtydzl.chromatic.com/?path=/story/stories-tooltip--default)


ryanwolhuter:
> is it possible to ignore mouse click while its opening? if you click the source of the tooltip it ends up dismissing it before you can see it. 

@daywiss this is the library's default behavior. I think this would be possible, but this lib is implemented according to the [aria design pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tooltip) so this current behavior is most likely what users expect

daywiss (approved):

PR #70, opened 2023-04-04
fix: fix settle and other issues

daywiss:
## motivation
settle button not showing up for verify tab which were expired.

## changes
testing this feature, ended up finding some more critical bugs:
1. updates to query state were not being reflected in panel - changed how panel references query state
2. once this was fixed, transactions in the action hook would loop infinitely, due to watching query state with transaction data, removed this dependency, but this may have unknown consequences
3. merged data from receipt would override previous data even if undefined, this was due to converters always exporting keys even when the values were undefined, causing overrides.  refactored converters to only return keys where data exists

finally the button is enabled now to "settle" when a vote had gone through liveness
image

### DIFF
--- a/src/contexts/OracleDataContext.tsx
+++ b/src/contexts/OracleDataContext.tsx
@@ -1,3 +1,4 @@
+import unionWith from "lodash/unionWith";
 import { config } from "@/constants";
 import {
   assertionToOracleQuery,
@@ -103,6 +104,22 @@ type ProcessAssertionsAction = DispatchAction<"assertions", Assertions>;
 
 type DispatchActions = ProcessRequestsAction | ProcessAssertionsAction;
 
+function mergeData(
+  prev: OracleQueryUI | undefined,
+  next: OracleQueryUI
+): OracleQueryUI {
+  // we must merge data in more information, since the next data may be mission previously queried data
+  const moreInformation = unionWith(
+    prev?.moreInformation ?? [],
+    next?.moreInformation ?? [],
+    (a, b) => a.title === b.title
+  );
+  return {
+    ...(prev || {}),
+    ...next,
+    moreInformation,
+  };
+}
 function DataReducerFactory<Input extends Request | Assertion>(
   converter: (input: Input) => OracleQueryUI
 ) {
@@ -113,10 +130,7 @@ function DataReducerFactory<Input extends Request | Assertion>(
     const { all = {} } = state;
     updates.forEach((update) => {
       const queryUpdate = converter(update);
-      all[update.id] = {
-        ...(all[update.id] || {}),
-        ...queryUpdate,
-      };
+      all[update.id] = mergeData(all[update.id], queryUpdate);
     });
     const init: {
       verify: OracleQueryList;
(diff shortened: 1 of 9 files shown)


src/contexts/OracleDataContext.tsx
```diff
@@ -103,6 +104,22 @@ type ProcessAssertionsAction = DispatchAction<"assertions", Assertions>;
 
 type DispatchActions = ProcessRequestsAction | ProcessAssertionsAction;
 
+function mergeData(
```
daywiss, line 107: this is so we can properly merge more information, need to dedupe and union the values from previous + new state

src/hooks/actions.tsx
```diff
@@ -52,15 +52,17 @@ export function usePrimaryPanelAction({
   const disputeAssertionAction = useDisputeAssertionAction({ query });
   // TODO: work out settle logic. need to clarify how we can detect when to show settle, and interactions
   // between change chain action, settle action and showing settled when on wrong chain
-  // const settlePriceAction = useSettlePriceAction({ query });
-  // const settleAssertionAction = useSettleAssertionAction({ query });
+  const settlePriceAction = useSettlePriceAction({ query });
+  const settleAssertionAction = useSettleAssertionAction({ query });
```
daywiss, line 56: enabling settle button

src/hooks/actions.tsx
```diff
@@ -238,7 +240,8 @@ export function useProposeAction({
           ]?.updateFromTransactionReceipt(receipt);
       })
       .catch(console.error);
-  }, [proposePriceTransaction, chainId, query]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [proposePriceTransaction]);
```
daywiss, line 244: removing query and chain to prevent infinite loop in hook. this would keep emitting notifications after tx completed, because this would then trigger a query update

src/types/ui.ts
```diff
@@ -75,7 +75,7 @@ export type OracleQueryUI = {
   queryText?: string;
   // for price requests the value text is null until a price is proposed. Then it is the proposed price. After a price is settled it is the settled price.
   // for assertions the value text is null until settlement, after which it is the `settlementResolution` field
-  valueText: string | null;
+  valueText?: string | null;
```
daywiss, line 78: more data that may not exist

src/helpers/converters.ts
```diff
@@ -636,39 +649,103 @@ export function requestToOracleQuery(request: Request): OracleQueryUI {
...
+  const result: OracleQueryUI = {
+    id,
+    chainId,
+    oracleType,
+    oracleAddress,
+    chainName: getChainName(chainId),
+    actionType: getRequestActionType(state),
+    moreInformation: [],
+    project: "UMA",
+  };
```
daywiss, line 662: refactoring the returned result so that we dont include keys to data which is undefined

src/hooks/actions.tsx
ryanwolhuter: I should not have included the query and chain id here in the first place, so removing this is all good

ryanwolhuter (approved):
excellent work as always, well done

src/hooks/actions.tsx
daywiss: this wasnt you, i added this when i refactored

PR #71, opened 2023-04-05
fix: close panel on pop state

ryanwolhuter:
We want to close the panel when the user clicks the back button in the browser.

### DIFF
--- a/src/hooks/handleQueryInUrl.ts
+++ b/src/hooks/handleQueryInUrl.ts
@@ -12,7 +12,8 @@ export function useHandleQueryInUrl() {
   const { page } = usePageContext();
 
   useEffect(() => {
-    const hasQueryUrl = Object.keys(router.query).length > 0;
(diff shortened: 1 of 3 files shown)


src/hooks/handleQueryInUrl.ts
```diff
@@ -12,7 +12,8 @@ export function useHandleQueryInUrl() {
   const { page } = usePageContext();
 
   useEffect(() => {
-    const hasQueryUrl = Object.keys(router.query).length > 0;
+    const hasQueryUrl = window.location.search !== "";
```
ryanwolhuter, line 15: for some reason the nextjs router has [weird behavior when navigating around with queries in the url](https://github.com/vercel/next.js/issues/10875). That caused this `hasQueryUrl` to return true when it wasn't supposed to. Checking the window search params directly solves the issue.

daywiss (approved):
nice

PR #74, opened 2023-04-05
fix: panel infinite loop

ryanwolhuter:
We were running the message generating logic when the `content` changed. This is redundant and can lead to an infinite loop, because of how react stores references in memory (they can move around)


daywiss (approved):
working for me

PR #76, opened 2023-04-05
fix: ensure buttons stay disabled during tx confirm

daywiss:
## motivation
action button keeps enabling while tx is being mined

## changes
this is due to different timing between waiting on tx wtih ethers and the wagmi hook to watch tx hash.  we want to use the data returned from wagmi hook to disable button once tx is mined.  this required some changes to the dependent calls, as certain things needed to be reset when changing which query we are looking at.

### DIFF
--- a/src/hooks/actions.tsx
+++ b/src/hooks/actions.tsx
@@ -11,7 +11,10 @@ import {
   disputed,
   insufficientBalance,
   propose,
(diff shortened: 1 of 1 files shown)


ryanwolhuter (approved):
🔥 🔥 🔥

PR #79, opened 2023-04-05
handle entries number changing

ryanwolhuter:
Handle the case where the list of entries gets shorter, ensuring that the pagination component stays consistent with the updated entries.

- If the new entries list is empty, set the page number to 1 and update the entries to show.
- Calculate the new number of pages based on the updated entries and the current resultsPerPage.
- If the current page number is greater than the new number of pages, set the page number to the new last page, and update the entries to show.
- If the current page number is still valid, just update the entries to show without changing the page number.


daywiss (approved):
lgtm

PR #81, opened 2023-04-06
fix: use primary action title to determine if already settled

ryanwolhuter:
The request is already settled if the page is not /settled and the primaryAction title is either undefined or "Settled".


daywiss (approved):
working for me, but this does add a "settled" button where there didnt used to be one. not sure if that was intended, or even cauesd by this change 
image

that actually only happened when i settled a request, then clicked through the link to go to the settled page. when viewing requests directly from settled page this button doesnt show

PR #83, opened 2023-04-06
chore: upgrade non breaking dependencies

ryanwolhuter:
I've updated all of the dependencies to the latest versions and checked that nothing broke. I checked the release changelogs for breaking changes to be sure.

**Note:** Ethers was not upgraded because wagmi still depends on v5


.storybook/main.ts
```diff
@@ -1,7 +1,7 @@
 import { StorybookConfig } from "@storybook/nextjs";
 
 const storybookConfig: StorybookConfig = {
-  stories: ["../src/**/*.mdx", "../src/**/*.stories.@(js|jsx|ts|tsx)"],
+  stories: ["../src/**/*.stories.@(js|jsx|ts|tsx)"],
```
ryanwolhuter, line 4: removed because we don't have any .mdx file stories and storybook v7 shows a warning if we include this config without them

next.config.js
```diff
@@ -2,9 +2,6 @@ const withSvgr = require("next-plugin-svgr");
 
 /** @type {import('next').NextConfig} */
 const nextConfig = {
```
ryanwolhuter, line 4: issue is resolved upstream now

daywiss (approved):

PR #84, opened 2023-04-06
feat: backwards compatible urls

ryanwolhuter:
### Add support for the old dapp's style of links

@daywiss please feel free to take over and change and merge this PR as you like when I'm away.

I've simplified the logic in the new dapp a bit here — now we only use the names `transactionHash` and `eventIndex` for both assertions and requests. This also makes dealing with the old links easier.

I've added support for using either a hash and log index, or the request details. We won't be using the request details style in the new dapp, that's only for backwards compatibility.

I tested with this slug which came straight from the old dapp

`/request?transactionHash=0xca205df65fda9b4d952dda74bb0cc38004ed2ad6c6c01ecb7323ed1dcff72ffc&chainId=1&oracleType=OptimisticV2&eventIndex=376`

and this one that I constructed myself, I'm not so sure exactly how the old links look though so I might be wrong here

`/request?chainId=1&oracleType=OptimisticV2&requester=0x7034362c495552470aD86a3A86A57c16Fe5C82B3&timestamp=1675198247&identifier=DECENTRA-LIST&ancillaryData=0x446f20616c6c2050726f706f73656420416464726573736573206d65657420746865204c697374204372697465726961206174207468652074696d65206f662074686520707269636520726571756573743f204c6973742043726974657269613a204164647265737365732074686174206861766520766f74656420696e206f76657220383025206f6620554d412044564d20566f74657320746861742068617665207265736f6c76656420696e207468652070726576696f75732031383020646179732e2e20446563656e7472612d4c697374205265766973696f6e204944203d20312e20466f7220646972656374696f6e7320746f2066696e64207468652050726f706f736564204164647265737365732c2073656520496d706c656d656e746174696f6e2073656374696f6e206f6620554d49502d3136392e`


next.config.js
```diff
@@ -17,6 +17,15 @@ const nextConfig = {
       ],
     },
   },
+  async redirects() {
```
ryanwolhuter, line 20: redirect the old /request route

daywiss (commented):
seems to have intermittent issues with some links, will look into it: `/request?transactionHash=0x27f0bb5d2f0e2d7be3d2756bc461aee22eaee9aa48d3f2e5f2919512f2ec5eed&chainId=137&oracleType=Optimistic&eventIndex=472`

daywiss (approved):
looks like the missing requests are due to the graph query. will fix in new np. looks good though

PR #86, opened 2023-04-10
feat: add skinny oo transaction handlers

daywiss:
## motivation
we want to be able to handle skinny oo updates when user submits transactions

## changes
this basically enables skinny oo ethers service configuration, copies most of the oo v1 ethers service in, and updates it to work with skinny oo calls. this is untested since we have no goerli subgraph for skinny oo, but it shouldnt break anything


ryanwolhuter (approved):
👍

PR #93, opened 2023-04-11
fix: clicking in panel scrolls page to top

ryanwolhuter:
Remove a configuration for the scroll lock lib that caused this to happen. I am not sure why I even added it in the first place. I think it's a leftover from the original implementation in the voter dapp.


libs/src/utils.ts
```diff
@@ -193,8 +193,12 @@ export function isUnique<T>(
 }
 
 export function parseIdentifier(identifier: string | null | undefined): string {
```
ryanwolhuter, line 195: @daywiss this function errored when you pass in a regular utf8 string. Caused some trouble in storybook. I figured we might as well just let the function still work when you pass in a utf8 string since that's what we want to end up with. let me know if that doesn't make sense

src/components/Panel/Content.tsx
```diff
@@ -20,12 +20,7 @@ export function Content({ children, panelOpen, closePanel }: Props) {
   }, [panelOpen]);
 
   return (
-    <FocusOn
-      enabled={panelOpen}
-      onEscapeKey={closePanel}
-      preventScrollOnFocus
```
ryanwolhuter, line 26: this `preventScrollOnFocus` is what caused the scrolling to the top thing

daywiss (approved):
nice, working

PR #99, opened 2023-04-13
improve: refactor panel into components

ryanwolhuter:
# Refactor panel into components

The panel component has a lot of code in it, and its all jumbled up together. A lot of it doesn't need to know about each other, so it would be better if it was separated from the business logic side of things. There are also a lot of redundant styles that can be consolidated.

There's no Story here because I verified via chromatic that this refactor has not changed anything visual.

The logic is all the same, just split out into pieces to make it easier to reason about. I've added comments on things that require attention.


src/components/Panel/Actions/Actions.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const primaryAction = usePrimaryPanelAction({
+    query,
+    proposePriceInput,
+  });
+
+  const { address } = useAccount();
+  const { chain: connectedChain } = useNetwork();
+  const { page } = usePageContext();
+
+  const pageIsPropose = page === "propose";
```
ryanwolhuter, line 34: I added these booleans to make it easier to reason about these states. logic is the same

src/components/Panel/Panel.tsx
```diff
@@ -51,539 +12,29 @@ const errorBackgroundColor = addOpacityToHsla(red500, 0.05);
...
-      <br />
-      <MessageLink
-        href="https://docs.uma.xyz/developers/setting-custom-bond-and-liveness-parameters"
-        target="_blank"
-      >
-        Learn more
-      </MessageLink>
-    </>
-  );
+  const props = !!content
```
ryanwolhuter, line 20: now the panel's only job is to wrap the other stuff in a panel and render it. setting the props like I've done here lets us do the not undefined assertion in one place, and then treat the query as defined in all the children. simplifies things quite a bit while keeping the logic the same.

src/components/Panel/Actions/ActionDetails.tsx
```diff
@@ -0,0 +1,126 @@
...
+    </Link>
+  </>
+);
+
+// todo: @sean update copy
+const rewardInformation = (
+  <>
+    <p>
+      Lorem ipsum dolor sit, amet consectetur adipisicing elit. Illum, beatae.
+      Lorem ipsum dolor sit amet consectetur adipisicing elit. Quisquam, quod.
```
daywiss, line 73: There was a recent pr from sean for copy, may want to check if this was included. If not someone else will need to make change

src/components/Panel/Actions/ActionDetails.tsx
```diff
@@ -0,0 +1,126 @@
+import { Currency, InformationIcon } from "@/components";
+import type { OracleQueryUI } from "@/types";
+import styled from "styled-components";
+import { Link, Text } from "../style";
+
+export function ActionDetails({
+  bond,
```
daywiss, line 7: Very confused by putting "action" in all these components. Action should be reserved for things user can do to mutate something, but many of these seem just for display

src/components/Panel/Actions/ActionDetails.tsx
ryanwolhuter: I re-added it when I moved it 👍

src/components/Panel/Actions/ActionDetails.tsx
ryanwolhuter: This component is called action details because its part of the actions component (it shows the text about the actions). But since this component is already in a directory called Actions, I renamed it to be just `Details` instead to avoid confusion. Are there other components that also have confusing names? The only other one with action in the name is `Actions`, which is the component that holds all the actions the user can take with the contract. I think that name makes sense.

daywiss (approved):

PR #106, opened 2023-04-19
fix: search missing on mobile

ryanwolhuter:
The search bar component is not present in the filters mobile wrapper, which means the search bar does not show on mobile.

* Add the search bar component to the mobile wrapper

Screenshot 2023-04-19 at 09 49 39


daywiss (approved):
thanks for fixing this!

PR #108, opened 2023-04-19
fix: flash of loader when content is loaded

ryanwolhuter:
Use a function initializer for the is loading state, which is persisted between renders, to avoid re-rendering with the loading state as `true` on first render.


daywiss (commented):
what happens if thers a page with no queries? its possible the verify page could be empty

daywiss (approved):

PR #109, opened 2023-04-19
feat: parse results for projects

ryanwolhuter:
The projects that we specifically handle are defined as:

`export const projects = [
  "Unknown",
  "Polymarket",
  "Cozy Finance",
  "Across",
  "Sherlock",
  "OSnap",
] as const;`

Of these, Sherlock does not appear in the UI because of how their flow works (they do everything in a single transaction). 

OSnap only runs on OOv3, and therefore their assertions will never end up on the propose page and don't need to have inputs. I have however added some logic for parsing the description from the OSnap claim data.

Polymarket has logic that has been copied wholesale from the voter dapp.

Cozy Finance and Across both use a simple yes = 1 no = 0 logic and can therefore have the options:

`[
    { label: "Yes", value: "1", secondaryLabel: "1" },
    { label: "No", value: "0", secondaryLabel: "0" },
  ]`

### DIFF
--- a/src/components/CheckboxDropdown.tsx
+++ b/src/components/CheckboxDropdown.tsx
@@ -36,9 +36,9 @@ interface Props {
 export function CheckboxDropdown({ title, items, onCheckedChange }: Props) {
   return (
     <DropdownRoot modal={false}>
-      <DropdownTrigger>
+      <_Trigger>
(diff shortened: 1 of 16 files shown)


daywiss (approved):
looks like this is working well

PR #111, opened 2023-04-19
fix: add final fee to bond

daywiss:
## motivation
final fee was not included in bond, some people were not seeing they would be debited more than what was shown on ui

## changes
this makes sure final fee gets added to bond in all tables

before
image
after
image

### DIFF
--- a/src/components/OracleQueryTable/ProposeCells.tsx
+++ b/src/components/OracleQueryTable/ProposeCells.tsx
@@ -2,12 +2,12 @@ import type { OracleQueryUI } from "@/types";
 import styled from "styled-components";
 import { Currency } from "../Currency";
 import { TD, Text } from "./style";
+import { exists } from "@libs/utils";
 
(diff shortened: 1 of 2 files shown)


src/components/OracleQueryTable/ProposeCells.tsx
```diff
@@ -2,12 +2,12 @@ import type { OracleQueryUI } from "@/types";
...
 import { Currency } from "../Currency";
 import { TD, Text } from "./style";
+import { exists } from "@libs/utils";
 
 export function ProposeCells({ query }: { query: OracleQueryUI }) {
   const { oracleType, tokenAddress, chainId, bond, reward } = query;
 
-  const hasReward = reward !== null;
-
+  const hasReward = exists(reward);
```
daywiss, line 10: reward could be undefined, exists checks null or undefined

ryanwolhuter (approved):

PR #112, opened 2023-04-20
fix: keep fetching requests until we have them all

ryanwolhuter:
The graph limits how many items we can request in a query. We need to keep querying and use the `skip` parameter to get the next batch. We can't do this with pages because we need _all_ the requests for stuff like getting old requests from the URL. I considered downloading all these older requests and serving them in a file, but many of the old requests are not resolved and therefore can change.

* Add `fetchAllRequests` and `fetchAllAssertions` logic that downloads all the requests/assertions by paging the data.


daywiss (commented):
@ryanwolhuter thanks for working on this, i want to update this to emit events as they come rather than wait to load all before and then emit a big batch as this could slow down initial loading

PR #113, opened 2023-04-20
fix: add ignore location to fuse config

ryanwolhuter:
The Fuse fuzzy search library has a default setting where it wants the search string to appear at the start of the search text. This is not what we want because we want to be able to find things anywhere in the string. By setting `ignoreLocation` to true, the search looks throughout the whole string instead.


daywiss (approved):
nice

PR #116, opened 2023-04-20
feat: parse descriptions as markdown

ryanwolhuter:
Use ReactMarkdown for the description field like in the voter dapp.

I've changed the type of `description` to just be a string, because only strings can be parsed as markdown. This meant I needed to change the osnap description to use markdown instead of a react element. I think this is actually much better as working with markdown in a context like that is easier than throwing around tsx all over the place.

Screenshot 2023-04-20 at 15 36 29


daywiss (approved):
awesome, looks like its working well.

PR #117, opened 2023-04-25
fix: add usePagination hook to avoid wrong initial values

ryanwolhuter:
### Fix: it takes long to go from settled to propose/verify

This bug was caused by this logic:

`const [rowsToShow, setRowsToShow] = useState(rows);`

Using the `rows` as the initial state means that the component is first rendered with the entire list of unpaginated values. This first render happens before the `Pagination` component has a chance to render, which means that we attempt to render all several thousand of the entries on the first render.

Changing the code to:

``const [rowsToShow, setRowsToShow] = useState(rows);``

Fixes the issue. 

However, this component's api is in my opinion going to be a source of bugs, because the `entriesToShow` logic is duplicated in the components that use it, and its certainly possible that it would happen again in another usage of the component.

I've opted to change the api of the component, so that its now intended to be used with a `usePagination` hook which holds all the state (including the state that was previously being duplicated in component implementations) and returns all of the props for the `Pagination` component. I think this api is much cleaner, and it wraps up the code that was a source of bugs and stops it from being duplicated.

I chose to put the `usePagination` hook in the same file as the `Pagination` component because they are only intended to be used together. If this doesn't make sense, I'm happy to move it to the `hooks` directory with the other hooks.

The **[Stories](https://63d101d13121a53332f6218c-zqfjbovbxx.chromatic.com/?path=/story/stories-pagination--even-number-of-entries)** introduce no visual changes and pass all tests.

### DIFF
--- a/src/components/OracleQueryList/OracleQueryList.tsx
+++ b/src/components/OracleQueryList/OracleQueryList.tsx
@@ -26,7 +26,7 @@ export function OracleQueryList({
   isLoading,
   findQueryIndex,
 }: Props) {
-  const [itemsToShow, setItemsToShow] = useState(items);
+  const [itemsToShow, setItemsToShow] = useState<typeof items>([]);
 
   useEffect(() => {
     if (items.length <= defaultResultsPerPage) {
--- a/src/components/OracleQueryList/OracleQueryList.tsx
+++ b/src/components/OracleQueryList/OracleQueryList.tsx
@@ -1,9 +1,8 @@
 import { defaultResultsPerPage } from "@/constants";
 import type { OracleQueryUI } from "@/types";
 import type { PageName } from "@shared/types";
-import { useEffect, useState } from "react";
 import styled from "styled-components";
-import { Pagination } from "../Pagination";
+import { Pagination, usePagination } from "@/components";
 import { Item } from "./Item";
 import { LoadingItem } from "./LoadingItem";
 
@@ -19,6 +18,7 @@ interface Props {
  * @param page The page the queries are being rendered on.
  * @param items The queries to render.
  * @param isLoading Whether the queries are still loading.
+ * @param findQueryIndex - the index of the query from the url, if any
  */
 export function OracleQueryList({
   page,
@@ -26,13 +26,11 @@ export function OracleQueryList({
   isLoading,
   findQueryIndex,
 }: Props) {
-  const [itemsToShow, setItemsToShow] = useState<typeof items>([]);
+  const { entriesToShow, ...paginationProps } = usePagination(
+    items,
+    findQueryIndex
+  );
 
-  useEffect(() => {
-    if (items.length <= defaultResultsPerPage) {
-      setItemsToShow(items);
-    }
-  }, [items]);
   return (
     <Wrapper>
       <Title>Query</Title>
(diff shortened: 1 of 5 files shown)


src/components/OracleQueryList/OracleQueryList.tsx
```diff
@@ -19,20 +18,19 @@ interface Props {
...
+ * @param findQueryIndex - the index of the query from the url, if any
  */
 export function OracleQueryList({
   page,
   items,
   isLoading,
   findQueryIndex,
 }: Props) {
-  const [itemsToShow, setItemsToShow] = useState(items);
+  const { entriesToShow, ...paginationProps } = usePagination(
```
ryanwolhuter, line 29: this new api is much cleaner and avoids duplicating the code which lead to the bug this pr is intended to fix

src/components/OracleQueryList/OracleQueryList.tsx
```diff
@@ -44,18 +42,14 @@ export function OracleQueryList({
...
         </>
       )}
       {items.length > defaultResultsPerPage && !isLoading && (
         <PaginationWrapper>
-          <Pagination
-            entries={items}
-            setEntriesToShow={setItemsToShow}
-            findIndex={findQueryIndex}
-          />
+          <Pagination {...paginationProps} />
```
ryanwolhuter, line 52: The pagination component gets everything it needs from the hook

src/components/Pagination.tsx
```diff
@@ -8,24 +8,27 @@ import {
...
+ * An array of entries to show for the current page.
+ * All of the props needed to render the pagination component.
  */
-export function Pagination<Entry>({
-  entries,
-  setEntriesToShow,
-  findIndex,
-}: Props<Entry>) {
+export function usePagination<Entry>(entries: Entry[], findIndex?: number) {
+  const [entriesToShow, setEntriesToShow] = useState<typeof entries>([]);
```
ryanwolhuter, line 24: This hook now holds all the pagination state, including the `entriesToShow` state that was previously being re-implemented in components.

ryanwolhuter: This logic was the cause of the bug — previously the `useState` used the whole list of entries as its initial value, which caused the component to attempt to render all thousands of them before the effect hook in the pagination could run. Starting off with an empty array fixes this.

src/components/Pagination.tsx
```diff
@@ -44,10 +47,44 @@ export function Pagination<Entry>({
     lastPageNumber - 1 !== numberOfButtons;
   const buttonNumbers = makeButtonNumbers();
 
+  const getPageNumberOfItem = useCallback(
```
ryanwolhuter, line 50: I wrapped up these functions that get used in dependency arrays in `useCallback`. Usually this is not necessary, but when a function is used in a dependency array without `useCallback` it will cause the effect to fire on every render, because a function is a different object on each invocation of the component unless memoized like so.

src/components/Pagination.tsx
```diff
@@ -154,64 +167,110 @@ export function Pagination<Entry>({
     updateEntriesForPageNumber(number);
   }
 
-  function nextPage() {
+  function goToNextPage() {
```
ryanwolhuter, line 170: I also updated a couple of these function names just to be super clear about what they do

PR #118, opened 2023-04-26
fix: add years months weeks and days to time format

ryanwolhuter:
When we format the time remaining in the liveness progress bar, we were only handling hours, minutes and seconds. This made the progress bar say that it was going to end in 4 hours, when it really meant to say that it was going to end in 4 days and 4 hours.

There's no direct way to format the duration the way we do in the designs, so I've instead used the `date-fns` format function and then replaced the years, months etc. in the string instead. Not the cleanest, but gets the job done without being overly complicated.


mrice32 (approved):
LGTM!

PR #121, opened 2023-04-28
fix: fix search terms to match oo ui type

daywiss:
## motivation
some search terms are not working as expected, such as searching for text within a description

## changes
the bug was due to having wrong property names in the search keys. this has been updated to match the oracle query type
image


src/constants/filters.ts
```diff
@@ -1,15 +1,21 @@
 export const keys = [
```
ryanwolhuter, line 1: I think we should add a type here to stop this from happening again. Doing so actually indicates that we have some other wrong fields here too.

```suggestion
export const keys: (keyof OracleQueryUI)[] = [
```

Adding the type shows that these keys are invalid:

"price",
 "formattedBond",
 "formattedReward",
 "assertion",

ryanwolhuter (approved):

src/constants/filters.ts
daywiss: great idea!

PR #122, opened 2023-04-28
improve: prevent site locking from search

daywiss:
## motivation
if trying to search large strings, the site locks up on settled page since there are so many entries and long strings

## changes
this removes some redundant search fields which contain long strings, related to ancillary data, and adds a debounce for the input box so search is not executed on every change, it waits 500ms for user to stop typing.


ryanwolhuter (approved):
Good catch

PR #123, opened 2023-05-02
fix: prioritize in liveness queries on verify page

ryanwolhuter:
Update the sorting logic for queries on the `verify` page to be more useful by separating out `inLiveness` and `notInLiveness` queries.

If a query has a `disputeHash` property then it is considered "disputed" and is therefore not in liveness.
If a query's `livenessEndsMilliseconds` property is in the past then the query is not in liveness.

We separate the queries before sorting.

We sort the `inLiveness` queries by `livenessEndsMilliseconds` in descending order so that the ones that expire soonest are at the top.

We sort the `notInLiveness` by `timeMilliseconds` (same as for other pages) because since they are not in liveness, the time that the liveness ends is no longer relevant.

Finally, we combine these two arrays such that the whole list of in liveness queries is shown first, and the rest come after.

I've updated **[this story](https://63d101d13121a53332f6218c-zzghnewlse.chromatic.com/?path=/story/pages-verify--with-different-expirations)** to demonstrate this behavior.

### DIFF
--- a/src/helpers/util.ts
+++ b/src/helpers/util.ts
@@ -1,7 +1,7 @@
 import { mobileAndUnder, tabletAndUnder } from "@/constants";
 import type { OracleQueryList } from "@/contexts";
 import type { DropdownItem, OracleQueryUI } from "@/types";
-import { capitalize, words } from "lodash";
+import { capitalize, orderBy, partition, words } from "lodash";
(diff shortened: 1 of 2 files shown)


daywiss (approved):
great

PR #125, opened 2023-05-18
feat: handle graph failures in queries

ryanwolhuter:
Pass down an `addErrorMessage` function from the `ErrorContext` to the graphql queries so that they can show a global error in the event of a graph outage like the planned one coming up on Monday.

We don't know if the graph will simply throw a 500 error during this downtime, or if it will return an errors object like it normally would in the case of a query error. I've updated the functions to handle both cases.

### DIFF
--- a/libs/src/oracle-sdk-v2/services/oraclev1/gql/queries.ts
+++ b/libs/src/oracle-sdk-v2/services/oraclev1/gql/queries.ts
@@ -24,9 +24,9 @@ export async function getPriceRequests(
     return result;
   } catch (e) {
     addErrorMessage({
-      text: "TheGraph is experiencing downtime. Please use the legacy dapp while they rectify the issue",
+      text: "The Graph is experiencing downtime",
       link: {
-        text: "Legacy Dapp",
+        text: "Please use the Legacy Dapp",
         href: "https://legacy.oracle.uma.xyz",
       },
     });
--- a/libs/src/oracle-sdk-v2/services/oraclev1/gql/queries.ts
+++ b/libs/src/oracle-sdk-v2/services/oraclev1/gql/queries.ts
@@ -1,6 +1,7 @@
 import { chainsById } from "@shared/constants";
 import type {
   ChainId,
+  ErrorMessage,
   OOV1GraphEntity,
   OOV2GraphEntity,
   OracleType,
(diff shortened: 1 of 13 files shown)


libs/src/oracle-sdk-v2/services/oraclev1/gql/queries.ts
```diff
@@ -12,13 +13,25 @@ import request, { gql } from "graphql-request";
...
-  return result;
+  try {
+    const chainName = chainsById[chainId];
+    const queryName = makeQueryName(oracleType, chainName);
+    const isV2 = oracleType === "Optimistic Oracle V2";
+    const result = await fetchAllRequests(url, queryName, isV2);
+    return result;
+  } catch (e) {
+    addErrorMessage({
+      text: "TheGraph is experiencing downtime. Please use the legacy dapp while they rectify the issue",
```
smb2796, line 27:
```suggestion
      text: "The Graph is experiencing downtime. Please use the legacy dapp.,
```

libs/src/oracle-sdk-v2/services/oraclev3/gql/queries.ts
```diff
@@ -1,13 +1,33 @@
...
+  addErrorMessage: (message: ErrorMessage) => void
+) {
+  try {
+    const chainName = chainsById[chainId];
+    const queryName = makeQueryName("Optimistic Oracle V3", chainName);
+    const result = await fetchAllAssertions(url, queryName);
+    return result;
+  } catch (e) {
+    addErrorMessage({
+      text: "TheGraph is experiencing downtime. Please use the legacy dapp while they rectify the issue",
```
smb2796, line 23:
```suggestion
      text: "The Graph is experiencing downtime. Please use the legacy dapp",
```

libs/src/oracle-sdk-v2/services/oraclev1/gql/queries.ts
```diff
@@ -12,13 +13,25 @@ import request, { gql } from "graphql-request";
...
+      link: {
+        text: "Legacy Dapp",
```
smb2796, line 29: Probably would be better to just hyperlink the first "legacy dapp" in the message rather than repeating.

libs/src/oracle-sdk-v2/services/oraclev3/gql/queries.ts
```diff
@@ -1,13 +1,33 @@
...
+      link: {
+        text: "Legacy Dapp",
```
smb2796, line 25: Probably would be better to just hyperlink the first "legacy dapp" in the message rather than repeating.

PR #127, opened 2023-06-09
fix: markdown in assertion claims shows syntax in assertion

ryanwolhuter:
If a claim is made using markdown syntax and they do not conform to a known pattern like with osnap, then we will use the first 100 characters to form the claim's title in the ui. However, if the claim uses markdown in those first 100 characters, we will see markdown syntax in the title. For example you would see something like "## My markdown title *bold" etc.

In future we could be smarter and for example parse out title indicators like # and use those as the claim title, but for now I think the fastest solution is to just strip out the markdown entirely and use the resulting non-syntax string as normal.
 
 * Add a `removeMarkdown` lib and use it to remove markdown in titles
 
 I've changed the claim text in all the oov3 mock data to include markdown, which you can see in this **[Story](https://63d101d13121a53332f6218c-smltrmbzgo.chromatic.com/?path=/story/pages-verify--default)**


daywiss (approved):
looks good

PR #129, opened 2023-07-14
feat: add web3 fallback for querying data if graph down

daywiss:
## motivation
we had a critical time where our subgraph was down, but we also needed to dispute a proposal. we need a way to query providers as a backup in case the graph goes down.

## changes
this adds a feature to each ethers service to query based on a number of blocks ago. it will automatically be called if any of the graphs throw an error during querying. appropriate error messages will be displayed if this happens, alerting the user the site experience may be degraded. 

image

## testing
this is difficult to test, you wil most likely need to checkout locally and edit your .env files to point to incorrect graph urls. this will trigger the fallback.

## todo
by default all lookback limits are set to 100000.  this may not be appropriate for all networks, if its a problem we can update these values per network and graph at a future point.

### DIFF
--- a/libs/src/oracle-sdk-v2/services/oracles/factory.ts
+++ b/libs/src/oracle-sdk-v2/services/oracles/factory.ts
@@ -1,4 +1,4 @@
-import type { ChainId, ErrorMessage, OracleType } from "@shared/types";
+import type { ChainId, OracleType } from "@shared/types";
 import type { Handlers, Service, ServiceFactory } from "../../types";
 
 // gql1 covers skinny, v1, v2
@@ -11,14 +11,12 @@ export type GqlConfig = {
   chainId: ChainId;
   type: OracleType;
   address: string;
-  addErrorMessage: (message: ErrorMessage) => void;
 };
 export type Config = GqlConfig[];
 
-export const Factory =
-  (config: Config): ServiceFactory =>
-  (handlers: Handlers): Service => {
-    const services: Service[] = config.map((config) => {
+export const Factory = (config: Config): ServiceFactory[] => {
+  return config.map((config) => {
+    return (handlers: Handlers): Service => {
       if (
         config.source === "gql" &&
         (config.type === "Optimistic Oracle V1" ||
@@ -33,15 +31,6 @@ export const Factory =
       throw new Error(
         `Unsupported oracle type ${config.type} from ${config.source}`
       );
-    });
-
-    async function tick() {
-      await Promise.all(
-        services.map(async (service) => (service ? service.tick() : undefined))
-      );
-    }
-
-    return {
-      tick,
     };
-  };
+  });
+};
--- a/libs/src/oracle-sdk-v2/services/oraclev1/ethers/factory.ts
+++ b/libs/src/oracle-sdk-v2/services/oraclev1/ethers/factory.ts
@@ -1,7 +1,12 @@
 import Events from "events";
 import { ethers } from "ethers";
 import { assertAddress } from "@shared/utils";
-import { parseIdentifier } from "@libs/utils";
+import {
+  parseIdentifier,
+  rangeStart,
+  rangeSuccessDescending,
+  rangeFailureDescending,
+} from "@libs/utils";
 import type {
   Request as SharedRequest,
   OracleType,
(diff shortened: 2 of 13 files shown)


libs/src/oracle-sdk-v2/services/oracles/factory.ts
```diff
@@ -11,14 +11,12 @@ export type GqlConfig = {
...
   address: string;
-  addErrorMessage: (message: ErrorMessage) => void;
 };
 export type Config = GqlConfig[];
 
-export const Factory =
-  (config: Config): ServiceFactory =>
-  (handlers: Handlers): Service => {
-    const services: Service[] = config.map((config) => {
+export const Factory = (config: Config): ServiceFactory[] => {
```
daywiss, line 17: this factory used to return a single "service" from multiple configurations. it now returns an array of services. this was required so we can know which configuration is erroring in the oracle context

libs/src/oracle-sdk-v2/services/oraclev1/ethers/factory.ts
```diff
@@ -147,11 +176,45 @@ export const Factory = (config: Config): [ServiceFactory, Api] => {
...
+    let rangeState = rangeStart({ startBlock, endBlock });
+    const { currentStart, currentEnd } = rangeState;
+    try {
+      await oo.update(currentStart, currentEnd);
+      rangeState = rangeSuccessDescending({ ...rangeState, multiplier: 1 });
+    } catch (err) {
+      rangeState = rangeFailureDescending(rangeState);
+    }
+  }
+  function queryLatestRequests(blocksAgo: number) {
```
daywiss, line 189: this is a new api call to each service which will cache events and emit any requests it finds to the oracle context

libs/src/oracle-sdk-v2/services/oraclev1/gql/factory.ts
```diff
@@ -9,25 +9,13 @@ export type Config = {
   chainId: ChainId;
   address: string;
   type: OracleType;
-  addErrorMessage: (message: ErrorMessage) => void;
```
daywiss, line 12: we dont need this callback. its handled through services emitting errors in oracle context

libs/src/oracle-sdk-v2/services/oraclev3/gql/queries.ts
```diff
@@ -1,33 +1,13 @@
...
-  chainId: ChainId,
-  addErrorMessage: (message: ErrorMessage) => void
-) {
-  try {
-    const chainName = chainsById[chainId];
-    const queryName = makeQueryName("Optimistic Oracle V3", chainName);
-    const result = await fetchAllAssertions(url, queryName);
-    return result;
-  } catch (e) {
-    addErrorMessage({
```
daywiss, line 22: again this is removed to prefer emitting this in the oracle context. we need to be able to check errors there in order to fallback to correct web3 service

src/contexts/OracleDataContext.tsx
```diff
@@ -186,17 +185,59 @@ export function OracleDataProvider({ children }: { children: ReactNode }) {
...
       assertions: (assertions) =>
         dispatch({ type: "assertions", data: assertions }),
       errors: setErrors,
     });
   }, []);
 
-  errors.forEach((error, i) => {
-    if (error) console.warn("Oracle data error", i, errors);
-  });
+  useEffect(() => {
```
daywiss, line 196: this contains the fallback logic.  if any service errors, we check which service it is and react appropriately.
if its a graph service, we initiate a call to the web3 fallback service.
if its a web3 fallback service, we know that we are probably missing data, and need to show user link to legacy site

src/contexts/OracleDataContext.tsx
```diff
@@ -186,17 +185,59 @@ export function OracleDataProvider({ children }: { children: ReactNode }) {
...
+      const web3Fallback =
+        oracleEthersApis?.[serviceConfig.type]?.[serviceConfig.chainId];
+      if (web3Fallback && web3Fallback.queryLatestRequests) {
+        const web3Config = config.providers.find(
+          (c) =>
+            c.type == serviceConfig.type && c.chainId == serviceConfig.chainId
+        );
+        web3Fallback.queryLatestRequests(
+          web3Config?.blockHistoryLimit ?? 100000
+        );
```
daywiss, line 222: this is where the magic happens

src/contexts/OracleDataContext.tsx
```diff
@@ -186,17 +185,59 @@ export function OracleDataProvider({ children }: { children: ReactNode }) {
...
+        // if we reach here, theres a subgraph thats not working, but we can still fetch limited  history through provider
+        addErrorMessage({
+          text: "The Graph is experiencing downtime, site may be slower than normal",
+        });
+      } else {
+        // if we reach here, we dont have a fallback for a specific subgraph, technically this shouldnt ever happen though
+        // if the app is configured correctly
+        console.warn({ serviceConfig, error });
+        addErrorMessage({
+          text: "Currently unable to fetch all data, check back later",
```
ryanwolhuter, line 232: I tried testing this by blocking api.thegraph.com in my browser's network tab. I also tried the way you suggested by changing the urls in my env to be wrong.

I'm getting the "Currently unable to fetch all data, check back later" error in my error banner, instead of the "The Graph is experiencing downtime, site may be slower than normal" one I expected.

The actual fetching seems to work well though 👍

ryanwolhuter (approved):
Approving so long  because the data fetching looks good 👍  just see my comment about the error message text

PR #130, opened 2023-07-18
fix: update to walletconnect v2

ryanwolhuter:
We have the same walletconnect issue on the oracle dapp as the voter dapp. Seems like no one has tried to use it with walletconnect recently, otherwise we'd have heard complaints.

We need to add a project id and the new required configs to wagmi for it to work.

* Upgrade rainbowkit and wagmi to latest patch releases (not the v1.x that uses viem)
* Created and added walletconnect project id (already added to env of all 3 projects on vercel)


daywiss (approved):
Nice thanks

PR #131, opened 2023-07-19
improve: immediately open panel in loading state when is url query

ryanwolhuter:
previously when you opened the app with a request in the url, it would look like nothing is happening. 

now when you open with a request url it will immediately open the panel and show the panel in a loading state with loading skeletons and spinners. this is better because the user knows that we are actually finding the request and that they must just hold on.

testing:

open the app with a request in the url


src/components/Panel/Panel.tsx
```diff
@@ -11,29 +15,77 @@ import { Title } from "./Title";
  * @see `PanelContext`
  */
 export function Panel() {
-  const { content, panelOpen, closePanel } = usePanelContext();
-
+  const { id, panelOpen, closePanel } = usePanelContext();
```
ryanwolhuter, line 18: now the panel is in charge of putting stuff in the url or taking it away.

src/hooks/handleQueryInUrl.ts
```diff
@@ -27,10 +27,42 @@ type SearchParams = HashAndIndexParams & RequestDetailsParams;
 
 export function useHandleQueryInUrl() {
   const router = useRouter();
-  const { openPanel, panelOpen } = usePanelContext();
+  const { setId, setPanelOpen } = usePanelContext();
   const { all: queries } = useQueries();
+  const [foundQuery, setFoundQuery] = useState<OracleQueryUI | undefined>();
   const { page } = usePageContext();
 
+  useEffect(() => {
```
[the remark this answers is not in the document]
ryanwolhuter: the panel can now be opened with no data. it will show loaders in that case

src/hooks/handleQueryInUrl.ts
```diff
@@ -27,10 +27,42 @@ type SearchParams = HashAndIndexParams & RequestDetailsParams;
 
 export function useHandleQueryInUrl() {
   const router = useRouter();
-  const { openPanel, panelOpen } = usePanelContext();
+  const { setId, setPanelOpen } = usePanelContext();
   const { all: queries } = useQueries();
+  const [foundQuery, setFoundQuery] = useState<OracleQueryUI | undefined>();
```
ryanwolhuter, line 32: we store the story in state so that various separate hooks can listen to it

daywiss (approved):
looks good, id suggest breaking out the refactors and logic changes to separate prs

PR #132, opened 2023-07-19
improve: logging and reduce gql query size

daywiss:
## motivation
we are seeing gql error banners frequently since we implemented fallback system. this is because we want to know when we are using web3 fallbacks. currently we dont know why the banners are showing up so much

## changes
this adds more general logging for any errors that happen when querying data. this will only log to console. after some inspection it looks like its the polygon subgraph failing, which may be due to the size of the queries. we also reduce the size to see if that helps prevent failures to the subgraph


ryanwolhuter (approved):
Nice

PR #134, opened 2023-07-21
chore: upgrade deps

ryanwolhuter:
In preparation for an upcoming task that we want to use the new `next/navigation` library for, which requires us to use the nextjs 13 /app directory setup, I've updated all of the project dependencies and migrated from the /pages dir to the /app dir.

The behavior of the app is unchanged.


.eslintrc.json
```diff
@@ -2,18 +2,15 @@
   "parser": "@typescript-eslint/parser",
   "parserOptions": {
     "tsconfigRootDir": ".",
-    "project": [
-      "./tsconfig.json"
-    ]
+    "project": true
```
ryanwolhuter, line 5: updated typescript-eslint and prettier, they now use slightly different configs

next.config.js
```diff
@@ -2,6 +2,10 @@ const withSvgr = require("next-plugin-svgr");
 
 /** @type {import('next').NextConfig} */
 const nextConfig = {
+  webpack: (config) => {
```
ryanwolhuter, line 5: this is some more crap from walletconnect. they seem to have somehow made the nextjs compiler think that these dependencies should be bundled in our project. see https://github.com/WalletConnect/walletconnect-monorepo/issues/1908

daywiss (approved):
lgtm

PR #135, opened 2023-07-21
improve: use tailwind

ryanwolhuter:
Install and configure tailwind
Use tailwind in some components that had weird behavior
Simplify how we handle svgs by leveraging tailwind


src/components/ChainNameAndIcon.tsx
```diff
@@ -1,31 +1,19 @@
...
 import type { ChainId, ChainName } from "@shared/types";
-import styled from "styled-components";
 
 interface Props {
   chainName: ChainName;
   chainId: ChainId;
 }
 export function ChainNameAndIcon({ chainName, chainId }: Props) {
-  const chainIcon = getChainIcon(chainId);
+  const ChainIcon = getChainIcon(chainId);
```
ryanwolhuter, line 9: with this new setup svgs can be used directly as components and they can be sized with tailwind classes

src/components/CheckboxList.tsx
```diff
@@ -55,7 +54,7 @@ const CheckboxesWrapper = styled.div`
   &:not(:last-child) {
     margin-bottom: 12px;
     padding-bottom: 16px;
-    border-bottom: 1px solid ${addOpacityToHsla(blueGrey700, 0.25)};
+    border-bottom: 1px solid ${addOpacityToColor("var(--blue-grey-700)", 0.25)};
```
ryanwolhuter, line 57: I've updated this function to use the new css `color-mix` property which means it works on any css color string including variables as seen here

src/components/LoadingSpinner.tsx
```diff
@@ -1,73 +1,50 @@
-import { darkText, red500 } from "@/constants";
-import type { CSSProperties } from "react";
-import styled, { keyframes } from "styled-components";
+import { addOpacityToColor } from "@/helpers";
```
ryanwolhuter, line 1: I used this loading spinner library in the tools dapp and it looks and works much better than my custom one. I added it here because the existing loading spinner was looking odd after installing tailwind for reasons I can't figure out

src/constants/icons.tsx
```diff
@@ -20,49 +21,49 @@ import Sherlock from "public/assets/icons/projects/sherlock.svg";
...
 
-export const projectIcons: Record<string, JSX.Element> = {
-  Unknown: <Unknown />,
-  "Cozy Finance": <Cozy />,
-  Polymarket: <Polymarket />,
-  Across: <Across />,
-  Sherlock: <Sherlock />,
-  OSnap: <OSnap />,
-  Rated: <Rated />,
+export const projectIcons = {
```
ryanwolhuter, line 24: typescript and nextjs have updated how they handle these svg as react components so that we don't need to type them like this anymore. this is great because it lets us use these svgs directly as components and pass classes to them

daywiss (approved):

PR #136, opened 2023-07-24
improve: use immer

ryanwolhuter:
In preparation for allowing filters based on the url, we need to simplify the filter and state logic. I have begun this process by using Immer to make the state mutations easier to reason about and safer. 

the app behavior remains unchanged.


daywiss (approved):
seems to be working

PR #138, opened 2023-07-27
improve: enrich rated assertions with data from their contract in ui

ryanwolhuter:
We should show the report data as acquired from the rated oracle contract in the panel.

* fetch the report and violations from the contract
* show this instead of the additional text data for these requests

Testing:

I can't seem to find the goerli transaction that they used as an example in the oracle ui. hardcoded that value ("5") just to see that everything is fetched and displayed correctly, and I can confirm that it is.

### DIFF
--- a/shared/constants/abi.ts
+++ b/shared/constants/abi.ts
@@ -14,41 +14,43 @@ export const proposePriceAbi = [
   },
 ];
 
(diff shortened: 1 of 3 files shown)


daywiss (approved):
great job

PR #140, opened 2023-07-28
feat: allow filters based on query strings

ryanwolhuter:
we want to be able to persist filter and search state across page renders, and also to be able to share filter and search states in urls.

add a new set of hooks that handle persisting the filter and search state in the url.

when adding a new filter or search term, populate the url with it.

allow multiple of the same filters to be checked by having multiple of the same query params.

when the page first loads, check the url for filter or search state and populate the filters and search accordingly.


src/components/Header/Nav.tsx
```diff
@@ -1,21 +1,24 @@
...
+  const searchParams = useSearchParams();
 
   return (
     <Wrapper>
       <NavItems>
         {navLinks.map(({ label, href }) => (
           <NavItem key={href}>
             <Link
-              href={href}
+              href={`${href}${
```
ryanwolhuter, line 19: pretty annoying that you have to do this. the new router treats search params as just another part of the url, which is true I guess. but that means that you have to explicitly keep the search params around when you navigate to a different url

libs/src/oracle-sdk-v1/store/store.ts
```diff
@@ -34,11 +37,11 @@ export default class Store<S> {
     this.emit(this.state, prevState);
   }
   // same as write
-  async writeAsync(cb: WriteCallbackAsync<S>): Promise<void> {
-    const prevState = this.state;
-    this.state = await produce(this.state, cb);
-    this.emit(this.state, prevState);
-  }
+  // async writeAsync(cb: WriteCallbackAsync<S>): Promise<void> {
```
ryanwolhuter, line 40: commented out because of type issue with the latest version of immer

src/components/Layout.tsx
```diff
@@ -7,14 +7,17 @@ import {
...
+import { useQueryInSearchParams } from "@/hooks/useQueryInSearchParams";
 import type { ReactNode } from "react";
 import { LegacyDappLinkBanner } from "./LegacyDappLinkBanner";
 import StyledComponentsRegistry from "./StyledComponentsRegistry";
+import { useFiltersInSearchParams } from "@/hooks/useFiltersInSearchParams";
 
 export function Layout({ children }: { children: ReactNode }) {
   const { page } = usePageContext();
-  useHandleQueryInUrl();
+  useQueryInSearchParams();
```
ryanwolhuter, line 19: this layout is only rendered once and is persisted across page changes. so it makes sense to do the hook invocations here

src/components/Panel/Actions/Actions.tsx
```diff
@@ -78,7 +76,7 @@ export function Actions({ query }: Props) {
   }
 
   return (
-    <Wrapper>
+    <div className="bg-grey-400 px-page-padding lg:px-7 pt-5 pb-6">
```
ryanwolhuter, line 79: just changed these styles to tailwind

src/components/Panel/Panel.tsx
```diff
@@ -13,29 +16,66 @@ import { Title } from "./Title";
...
+      {props ? (
         <>
           <Title {...props} />
           <Actions {...props} />
           <InfoIcons {...props} />
           <Details {...props} />
         </>
+      ) : (
+        <>
+          <Title isLoading={true} close={close} />
```
ryanwolhuter, line 45: this is the loading version of the panel. doesn't receive data or do anything else

src/components/Panel/RatedAssertionTextData.tsx
```diff
@@ -27,6 +27,7 @@ export function Report(props: Props) {
     abi: ratedAbi,
     address: props.chainId === 1 ? mainnetAddress : goerliAddress,
     functionName: "reports",
+    chainId: props.chainId,
```
ryanwolhuter, line 30: I realized that I should have included the chain ID here

src/components/Panel/style.tsx
```diff
@@ -1,63 +1,43 @@
...
-  display: flex;
-  align-items: center;
-  gap: 16px;
-  margin-top: 20px;
-  padding-inline: 16px;
-  background: ${addOpacityToColor("var(--red-500)", 0.05)};
-  border: 1px solid var(--red-500);
-  border-radius: 2px;
-`;
+import type { ComponentPropsWithoutRef } from "react";
```
ryanwolhuter, line 2: these are all the same, just styles converted to tailwind

src/hooks/useQueryInSearchParams.ts
```diff
@@ -0,0 +1,195 @@
...
+    case "OptimisticV2":
+      return "Optimistic Oracle V2";
+    case "Skinny":
+      return "Skinny Optimistic Oracle";
+    default:
+      return undefined;
+  }
+}
+
+type State = {
```
ryanwolhuter, line 32: this is where we handle opening the panel when there is a query in the url

src/hooks/useUrlBar.ts
```diff
@@ -0,0 +1,87 @@
+import { usePathname, useRouter, useSearchParams } from "next/navigation";
+import { useCallback, useMemo } from "react";
+
```
ryanwolhuter, line 3: this hook provides an interface for interacting with the url

src/hooks/useUrlBar.ts
```diff
@@ -0,0 +1,87 @@
...
+      for (const [key, value] of Object.entries(params)) {
+        urlSearchParams.set(key, value);
+      }
+
+      router.push(`${pathname}?${urlSearchParams.toString()}`);
+    },
+    [pathname, router, searchParams],
+  );
+
+  const removeSearchParam = useCallback(
```
ryanwolhuter, line 40: super annoyingly, this is the only way to remove a param with a given name and value.

the api is supposed to support doing this https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/delete

but its only available in firefox at the moment.so we need this workaround

daywiss:
looks good. only thing im seeing wrong is some weird back/forward behavior, it seems to be ignored for both filters/searches and selecting queries. it looks like back behavior in production is also broken, so we can probably just address this as separate issue.

daywiss (approved):
thanks for going through pr with me, looks great. 

i think main learning from this would be to move source of truth for url state to an object, as it would make it much easier to manipulate, especially with immer, and then sync that as it changed to url or other parts of app. no need to change now though since you wrote some utilities

some weird back/forward issues, but its not clear how those should actually work, we can change in new pr

PR #141, opened 2023-08-01
fix: dropdown z index

ryanwolhuter:
When I converted the panel to tailwind, I overlooked an implementation detail: for reasons I'm not so sure about, [tailwind specifies z indexes in multiples of 10](https://tailwindcss.com/docs/z-index#setting-the-z-index).

In the rest of the app, we count z index in increments of 1. I had the panel set to z index of 1, and the dropdown to a z index of 2.

This meant that when I changed the panel to tailwind and gave it tailwind's equivalent of z  index 1, which is actually z index 10, it broke the implementation by making it a higher z index than that of the dropdown and caused the dropdown to be hidden behind the panel.

I missed this because I did not have a storybook example with the dropdown input for the propose state, which has now been rectified.

I have updated all the other z indexes used in the app to the max value, since each of those items is not intended to be used above one another, and also they are always expected to appear on top of everything else: the tooltips and the notifications.

Going forward these issues will be avoided entirely in future implementations — we now have the native [html dialog element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) for this, and it puts the items in a special "top" layer that is always above everything else regardless of z index.

### DIFF
--- a/src/components/Notifications.tsx
+++ b/src/components/Notifications.tsx
@@ -42,7 +42,7 @@ export function Notifications() {
 const AnimatedNotification = animated(Notification);
 
 const Wrapper = styled.div`
-  z-index: 2;
+  z-index: 9999;
   position: fixed;
   top: 0;
   left: 0;
--- a/src/components/RadioDropdown.tsx
+++ b/src/components/RadioDropdown.tsx
@@ -57,7 +57,6 @@ const _Trigger = styled(DropdownTrigger)`
 
 const _Content = styled(DropdownContent)`
   padding-block: 0;
-  z-index: 2;
 `;
 
 const _RadioItem = styled(RadioItem)`
--- a/src/components/Tooltip.tsx
+++ b/src/components/Tooltip.tsx
@@ -71,7 +71,7 @@ const slideLeftAndFade = keyframes`
 `;
 
 const Content = styled(RadixTooltip.Content)`
-  z-index: 2;
+  z-index: 9999;
   max-width: 400px;
   overflow-wrap: break-word;
   padding: 20px;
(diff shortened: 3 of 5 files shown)


src/components/RadioDropdown.tsx
```diff
@@ -57,7 +57,6 @@ const _Trigger = styled(DropdownTrigger)`
 
 const _Content = styled(DropdownContent)`
   padding-block: 0;
```
ryanwolhuter, line 59: remove override and use base value which is the max value

src/components/style.tsx
```diff
@@ -48,6 +48,7 @@ export const DropdownContent = styled(RadixDropdown.Content)`
   background: var(--white);
   border: 1px solid var(--blue-grey-400);
   border-radius: 4px;
+  z-index: 9999;
```
ryanwolhuter, line 51: dropdowns should always be the top layer

src/components/Notifications.tsx
```diff
@@ -42,7 +42,7 @@ export function Notifications() {
 const AnimatedNotification = animated(Notification);
 
 const Wrapper = styled.div`
-  z-index: 2;
+  z-index: 9999;
```
ryanwolhuter, line 45: Notifications should always be the top layer

src/components/Tooltip.tsx
```diff
@@ -71,7 +71,7 @@ const slideLeftAndFade = keyframes`
 `;
 
 const Content = styled(RadixTooltip.Content)`
-  z-index: 2;
+  z-index: 9999;
```
ryanwolhuter, line 74: Tooltips should always be the top layer

daywiss (approved):
nice find

PR #148, opened 2023-09-05
feat: add snapshot space links

daywiss:
## motivation
we want more information available for osnap requests

## changes
this adds some additional links, link to ipfs json data, and snapshot proposal space based on that data. it wont load for older formatted ipfs links since they did not include proposal information.

image


src/components/Panel/OSnapTextData.tsx
```diff
@@ -0,0 +1,59 @@
...
+            {snapshotLink.title}
+          </Link>
+        </div>
+      ) : undefined}
+      <AdditionalTextData {...props} />
+    </>
+  );
+}
+
+function snapshotProposalLink(ipfsData: unknown) {
```
ryanwolhuter, line 32:
```suggestion
type IpfsData = {
  data: {
    message: {
      title: string;
      space: string;
    };
  };
  hash: string;
};

function snapshotProposalLink(ipfsData: Partial<IpfsData>) {
  const title = ipfsData?.data?.message?.title;
  const space = ipfsData?.data?.message?.space;
  const hash = ipfsData?.hash;

  if (space && hash) {
    return {
      link: `https://snapshot.org/#/${space}/proposal/${hash}`,
      // eslint-disable-next-line
      title: title ?? space,
    };
  }
}

async function getIpfs(hash?: string) {
  if (!hash) return undefined;
  const response = await fetch(`https://snapshot.4everland.link/ipfs/${hash}`);
  return (await response.json()) as Partial<IpfsData>;
}

export function useIpfs(hash?: string) {
  return useSWR(`/ipfs/${hash}`, () => {
    assert(hash, "Missing ipfs hash");
    return getIpfs(hash);
  });
}
```

ryanwolhuter (approved):
lgtm

PR #149, opened 2023-09-08
fix: dont show disputed when settle is available

daywiss:
## motivation
theres a bunch of expired proposals that cant be settled on testnet for example: https://testnet.oracle.umaproject.org/?chain=5&transactionHash=0x95970c9252a4e1254aba892c81556d613b39e7c8fa92b365d00a69e20a226f5f&eventIndex=10

## changes
we were not correctly gaurding the dispute function when action type was set to "settle".  this exits so settle code can correctly take over


ryanwolhuter (approved):
🙏🏻

PR #150, opened 2023-10-02
feat: add polymarket bulletin updates

daywiss:
## motivation
Polymarket has a feature where they can dynamically give more context to polymarket questions through an on chain mechanism. We want to display that in the UI

## changes
Added polymarkets bulletin board feature: 
image
which requires querying their bulletin contract based on parameters we can fish out of the ancillary data.


src/helpers/util.ts
```diff
@@ -207,3 +208,15 @@ export function isValidOracleType(
...
   return !!oracleType && oracleType in oracleTypes;
 }
+
+export function isWagmiAddress(
+  maybeAddress: string | undefined,
+): maybeAddress is Address {
+  if (!maybeAddress) return false;
+  return "0x" == maybeAddress.slice(0, 2);
+}
+export function assertWagmiAddress(maybeAddress: string): Address {
```
ryanwolhuter, line 218: no need to change anything but you might find this cool https://blog.logrocket.com/assertion-functions-typescript/

you can annotate this type of function like `asserts value is Address` and it will automatically treat the value as `Address` after. saves some typecasting

ryanwolhuter (approved):
lgtm

PR #153, opened 2023-11-16
fix: show yes no dropdown for yes no query

daywiss:
## motivation
there are some yes no query identifiers not showing the right  option selections in details page

## changes
this adds a simple yes/ no dropdown if it detects a yes no identifier


gsteenkamp89 (approved):
I don't have that much context here but this looks food to me

PR #154, opened 2023-12-04
replace unhelpful "cannot estimate gas" error message

gsteenkamp89:
closes #UMA-1934

## Motivation

Wagmi egnerally does a good enough job of parsing/decoding any revert or error messages that might be thrown while attempting to interact with a contract. But in some cases there is either no sighash or string associated with the revert and in those cases the provider throws an "Internal JSON RPC error" along with the stringified blob of transaction data. 

This is a huge chuck of text and is not helpful to the user since all it's saying it "Cannot estimate gas" because we the provider was unable to simulate the transaction successfully. 

In these cases we want to simply replace this chunk of text with a generic message like "Transaction Failed".


daywiss (approved):
thanks for looking into this

daywiss:
you need to run prettier to get ci to pass

PR #156, opened 2023-12-14
feat: allow commas in p0 & p1 res

gsteenkamp89:
The app now parses p1/p2 proposals (no p3) and shows a dropdown in this case.

## screenshots
Screenshot 2023-12-14 at 17 05 54


daywiss (approved):
thanks!

PR #157, opened 2023-12-27
fix: ensure neg risk markets arent always shown by making it more spe…

daywiss:
## motivation
neg risk markets were being shown where they werent supposed to be ( missing options )

## changes
this makes the string to search for neg risk markets more specific by ending with period. also adds comments around why we have other specific strings. also removes "early request" as that is more like a voting result than a yes no proposal.

### DIFF
--- a/src/helpers/queryParsing.ts
+++ b/src/helpers/queryParsing.ts
@@ -231,20 +231,24 @@ function dynamicPolymarketOptions(
 export function maybeMakePolymarketOptions(
   decodedAncillaryData: string,
 ): DropdownItem[] | undefined {
+  // this is a specific search to look for a misspelling with options "p2 to a Yes"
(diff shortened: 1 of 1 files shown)


gsteenkamp89 (approved):
Looks good! 
Although you might've just swapped options1 and options2 around so you could simplify this i think

daywiss:
> Looks good! Although you might've just swapped options1 and options2 around so you could simplify this i think

yea originally i thought that, but the problem is that these options defined are actually special cases, and if none of these get a match it needs to fallback to the "dynamic options".  the problem with this neg risk one is its so short it will match with a lot of the  proposals that would have otherwise fallen through to dynamic parsing, so it needed its own special case that wouldnt accidentally catch a bunch of other requests

PR #158, opened 2024-02-14
feat: add core network

Reinis-FRP:
Fixes: https://linear.app/uma/issue/UMA-2396/add-core-to-oo-app

Adds core mainnet. We still would need to update rainbowkit dependencies in order to allow connect to core and post proposals/disputes.


daywiss (approved):
looks great

PR #161, opened 2024-02-27
feat: migrate to wagmi and rainbowkit 1.0 to enable core

daywiss:
## motivation
we want to add core, it required bumping some dependencies on wallet versions

## changes
this bumps wagmi/rainbowkit to v1 which is still behind the current latest version at 2. This required less changes than full migration to 2 while still being able to support core.


tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "ES2015",
+    "target": "ES2020",
```
daywiss, line 3: needed to support bigint notation

package.json
```diff
@@ -24,32 +24,8 @@
     "@radix-ui/react-portal": "^1.0.3",
     "@radix-ui/react-progress": "^1.0.3",
     "@radix-ui/react-tooltip": "^1.0.6",
-    "@rainbow-me/rainbowkit": "^0.12.18",
+    "@rainbow-me/rainbowkit": "^1.0.0",
```
daywiss, line 27: could probably play with this version ( bump higher ) , but it seems to be ok

package.json
```diff
@@ -66,25 +42,49 @@
...
+    "styled-components": "^6.0.5",
+    "superstruct": "^1.0.3",
+    "swr": "^2.2.0",
     "tailwindcss": "^3.3.3",
     "ts-node": "^10.9.1",
-    "typescript": "5.1.6"
+    "typescript": "5.1.6",
+    "use-immer": "^0.9.0",
+    "usehooks-ts": "^2.9.1",
+    "wagmi": "^1.4.13"
```
daywiss, line 87: this is the first version where they added core to list of chains ( looked up the blame )

shared/types/graphql.ts
```diff
@@ -385,15 +384,15 @@ export type ParsedOOV1GraphEntity = {
   time: string;
   requester: Address;
   currency: Address;
-  reward: BigNumber;
-  finalFee: BigNumber;
+  reward: bigint;
+  finalFee: bigint;
```
daywiss, line 388: the biggest pain is that in the version wagmi stopped using ethers and bignumbers, switched to viem and bigint, so everything needs to be converted

src/hooks/actions.tsx
```diff
@@ -281,12 +284,13 @@ export function useProposeAction({
       success: <>Proposed price</>,
       error: <>Failed to propose price</>,
     })
-      .then((receipt) => {
-        if (receipt && query)
-          oracleEthersApis?.[query.oracleType]?.[
-            chainId
-          ]?.updateFromTransactionReceipt(receipt);
-      })
+      // TODO: this is disabled because wagmi move from ethers, should look into updating this eventually
```
daywiss, line 287: wagmi switch from ethers broke interop between the fe and fetching services in libs which still use ethers

Reinis-FRP (approved):
great, thanks for pulling this together!

PR #162, opened 2024-03-01
fix: add skinny query params to allow settle

daywiss:
## motivation
we werent able to settle some across proposals. This was due to missing data on gql

## changes
This adds bond and customLiveness to the gql query for skinny data.  This enables the currently blocked across proposals from settling.  Accidentally linted the library folder as well which is why theres a bunch of other changes.


src/contexts/OracleDataContext.tsx
```diff
@@ -48,6 +48,7 @@ const [oracleEthersServices, oracleEthersApis] = config.providers
     if (config.type === "Optimistic Oracle V3")
       return [config, ...oracle3Ethers.Factory(config)];
     // skinny optimistic oracle is left
+    console.log("making skinny", config);
```
gsteenkamp89, line 51: Want to keep this log?

gsteenkamp89 (approved):
LGTM

PR #167, opened 2024-04-03
feat: add base envs and constants

daywiss:
## motivation
we want to enable base in the ui

## changes
add required envs and constants.  envs have been added to vercel for the prod deployment.

### DIFF
--- a/shared/constants/oracle.ts
+++ b/shared/constants/oracle.ts
@@ -39,7 +39,7 @@ export const chainsById = {
   288: "Boba",
   416: "SX",
   1116: "Core",
-  8353: "Base",
+  8453: "Base",
   43114: "Avalanche",
   42161: "Arbitrum",
   80001: "Mumbai",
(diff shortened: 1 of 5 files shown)


shared/constants/oracle.ts
```diff
@@ -39,6 +39,7 @@ export const chainsById = {
   288: "Boba",
   416: "SX",
   1116: "Core",
+  8353: "Base",
```
Reinis-FRP, line 42:
```suggestion
  8453: "Base",
```

shared/utils/ethers.ts
```diff
@@ -151,6 +153,8 @@ export function getBlockExplorerNameForChain(chainId: ChainId) {
       return "SX Explorer";
     case 1116:
       return "Core Scan";
+    case 8353:
```
Reinis-FRP, line 156:
```suggestion
    case 8453: 
```

src/constants/env.ts
Reinis-FRP, line 1: Also need adding 8453 to `const ChainId`

Reinis-FRP (approved):

PR #170, opened 2024-04-16
feat: show migration banner if using legacy plugin

gsteenkamp89:
## motivation

We want to let oSnap users know that they are missing out on Tenderly Sim functionality if they are using the legacy safeSnap plugin.

## screenshots

Screenshot 2024-04-16 at 16 36 35


daywiss (approved):
great!

PR #173, opened 2024-04-22
fix: prevent unbroken strings from overflowing error container

gsteenkamp89:
## motivation

Any unbroken strings (like addresses) would overflow its container. this is default css.
Here we ensure to break long words to prevent this from happening.


daywiss (approved):
looks better, thanks!

PR #174, opened 2024-04-29
fix: parsing early values incorrectly

gsteenkamp89:
## motivation

The app would not find "Early request" as an option in the ancillary data and therefore would always show the magicNumber as the settled value, instead of the corresponding label of "Early request".

## NOTE:
this [commit](https://github.com/UMAprotocol/optimistic-oracle-dapp-v2/pull/174/commits/f65b91133812fa8dcbb2bbc6a32c2e331d44bfbe) is the fix, the rest is formatting changes due to disabling eslint's formatting rules.
## Work done

- ESlint and Prettier are both attempting to format code. This PR removes eslint's formatting errors so we only use prettier.
- Correctly parse early request magic number


src/helpers/util.ts
```diff
@@ -168,6 +169,10 @@ export function maybeGetValueTextFromOptions(
   valueText: string | null | undefined,
   options: DropdownItem[] | undefined,
 ) {
+  //  early request not in original proposal
+  if (isEarlyVote(valueText)) {
+    return "Early Request";
+  }
```
gsteenkamp89, line 175: we will never find the magic number in options so we just check it here.

src/constants/earlyRequestMagicNumber.ts
```diff
@@ -1,9 +1,24 @@
...
+
+  const asBigInt = takeIntegerAsBigNumber(value)
+ return [earlyRequestMagicNumber, earlyRequestMagicNumberWei].map(takeIntegerAsBigNumber).some(num => num.eq(asBigInt))
 }
+
+function takeIntegerAsBigNumber(value: string) {
+  // remove commas, if present, remove decimals, take only integer
+  return bigNumberFromFloatString(value.replace(/,/g, "").split(".")[0])
+}
+
```
gsteenkamp89, line 24: only compare the integer part of the numbers, because the value we get from the proposal data is formatted (has comma separators and is only to a precision of 2)

daywiss (approved):
looks like it working, thanks

PR #180, opened 2024-06-06
use UTC date-time in polymarket bulletin board

gsteenkamp89:
fixes UMA-2611

## motivation
 It can be difficult to make voting decisions based on timing, when dates and times are displayed in different time zones. In this PR we use UTC time ion the Polymarket Bulletin board instead of local time.


daywiss (approved):
looks good, though couldnt find a way to see this in the site

PR #181, opened 2024-06-13
fix: update incomplete event data with direct reads from contract

gsteenkamp89:
fixes UMA-2644

## The bug

When a subgraph is down we resort to fetching request data directly from the chain. We do this first by fetching events from the OO, namely: `RequestPrice` , `ProposePrice`, `DisputePrice` & `Settle`.

For the OOv2 we SUM the bond and finalFee amounts. And when fetching dat like this ONLY rom events, we do not have the bond amount, only the final fee, thus causing teh bug that displays the full bond as only the final fee amount.

This PR ensure that, in the case of a subgraph outage, we first fetch event data, but also get updated data for each request by calling `getRequest()` on the OO.


Reinis-FRP (approved):
looking great!

PR #182, opened 2024-07-02
fix: don't disable proposal input if not connected or wrong chain

gsteenkamp89:
## To test

1. Go to propose page, open side panel by clicking on a query
2. make sure wallet in DISCONNECTED
3. See that the input is NOT DISABLED
4. Connect wallet.
5. Switch to incorrect chain.
6. See that the input is still NOT DISABLED


daywiss (approved):
awesome

PR #183, opened 2024-07-08
feat: format cryptic errors when proposing an already proposed request

gsteenkamp89:
closes UMA-2180 UMA-2659

## motivation

The oracle dapp displays some pretty large and unhelpful revert/error messages if a user attempts to propose/dispute/settle any assertions that have already been proposed/dispued/asserted.

This PR ensures we format the revert and error messages both in the tooltip (if button is disabled and preparation for the action failed) as well as in the red error container below the action panel in the case that the preparation succeeded so the user was able to submit a failing transaction.


daywiss (approved):
looks great

PR #184, opened 2024-07-19
add space name to osnap request titles

gsteenkamp89:
## motivation

We want users to be able to find osnap requests easier, at a glance. This PR extracts the space name from the request data and adds it to the title.

Screenshot 2024-07-19 at 14 04 47


daywiss (approved):
great

PR #187, opened 2024-07-30
show bond & reward tokens in block explorer

gsteenkamp89:
## motivation

Users would like to check the address of a token for a given chain. Bond & reward tokens in the Actions panel are now shown as links that direct to the relevant block explorer.

### screenshots

Screenshot 2024-07-30 at 20 34 00


daywiss (approved):
looks good

PR #190, opened 2024-09-13
improve: incremental loading for gql requests

daywiss:
# motivation
if visiting a specific request on polymarket for the first time, it can take a long time before it finally loads in.  timed it at 18 second.

# changes
Previously we were waiting for all requests to load in before updating the UI, now we load and update in batches allowing more recent request to come in first.  this allowed one to load in 6 seconds, compared to 18 seconds previously.


libs/src/oracle-sdk-v2/services/oraclev1/gql/queries.ts
```diff
@@ -20,6 +20,61 @@ export async function getPriceRequests(
   return result;
 }
 
+export async function* getPriceRequestsIncremental(
```
gsteenkamp89, line 23: actually a good use case for generator 👍

gsteenkamp89 (approved):
Nice this is a huge perf improvement! 👏 
LGTM

PR #192, opened 2024-09-25
feat: add individual request query for details page

daywiss:
# motivation
we want to speed up loading when directed to a single request by tx hash

# changes
this exposes a new function on oov2 ethers services to load something by a tx hash, currently only enabled on oov2 because polymarket needs the most optimization 

# testing
go to settled or proposed tab, choose a really old request, copy url and load it into an incognito window to see how fast it loads the details panel.  try this with the prod site to compare speed


gsteenkamp89 (approved):
looks good much faster

PR #193, opened 2024-09-26
defer action simulations until bond is approved

gsteenkamp89:
closes UMA-2773

- wait until bond is approved before we simulate any write functions that would revert if not approved

### DIFF
--- a/src/hooks/actions.tsx
+++ b/src/hooks/actions.tsx
@@ -59,9 +59,22 @@ export function usePrimaryPanelAction({
   // these are in order of importance
   const accountAction = useAccountAction({ query });
   const approveBondAction = useApproveBondAction({ query });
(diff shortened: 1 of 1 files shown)


src/hooks/actions.tsx
```diff
@@ -59,9 +59,22 @@ export function usePrimaryPanelAction({
...
+    proposePriceInput,
+    prepare: bondAmountApproved,
+  });
+  const disputeAction = useDisputeAction({
+    query,
+    prepare: bondAmountApproved,
+  });
+  const disputeAssertionAction = useDisputeAssertionAction({
+    query,
+    prepare: bondAmountApproved,
```
daywiss, line 76: waht does this prepare option do?

daywiss (approved):
tested this on sepolia with usdc, seems to work as expected!

PR #194, opened 2024-10-01
fix: hide expiration time if set to 0

daywiss:
# motivation 
highlighted area should not show if not in verify phase
image

# changes
if expiration time is set to 0, leave it undefined in data 
image


gsteenkamp89 (approved):
surgical 👍

PR #195, opened 2024-10-10
support infinite games

gsteenkamp89:
closes UMA-2784

Adds support for Infinite Games

Screenshot 2024-10-10 at 20 05 39


daywiss (approved):
ty

PR #197, opened 2024-10-31
add blast sepolia config

gsteenkamp89:
closes UMA-2789

### DIFF
--- a/src/constants/customChains.ts
+++ b/src/constants/customChains.ts
@@ -60,7 +60,7 @@ export const blast = defineChain({
 
 export const blastSepolia = defineChain({
   id: 168587773,
-  name: "Blast Spolia",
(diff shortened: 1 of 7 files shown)


src/constants/customChains.ts
```diff
@@ -57,3 +57,30 @@ export const blast = defineChain({
     },
   },
 });
+
+export const blastSepolia = defineChain({
+  id: 168587773,
+  name: "Blast Spolia",
```
daywiss, line 63: typo

daywiss (approved):
looks good minor typo

src/constants/customChains.ts
gsteenkamp89: good catch!

PR #200, opened 2024-11-06
fix: fix ethers based loading

daywiss:


src/contexts/OracleDataContext.tsx
```diff
@@ -195,6 +196,10 @@ export function OracleDataProvider({ children }: { children: ReactNode }) {
         dispatch({ type: "assertions", data: assertions }),
       errors: setErrors,
     });
+    oracleEthersApiList.map(
```
daywiss, line 199: this is a test. super important we never do this unless the graph call is missing for this chain/oracle.  that is not the easiest thing to figure out cuz the configs are so messy

gsteenkamp89 (approved):

PR #201, opened 2024-11-11
support story odyssey

gsteenkamp89:
closes UMA-2790

@daywiss 
The one and only AssertionMade event emitted by this OOv3 contract is currently over 200_000 blocks old. 
If you change the max lookback to 300_000, then the app does find it, update it from tx receipt etc, but for some reason does not show up in the UI!

### DIFF
--- a/libs/src/oracle-sdk-v2/services/oraclev2/ethers/factory.ts
+++ b/libs/src/oracle-sdk-v2/services/oraclev2/ethers/factory.ts
@@ -186,7 +186,6 @@ export const Factory = (config: Config): [ServiceFactory, Api] => {
     let rangeState = rangeStart({ startBlock, endBlock, maxRange: 10000 });
     do {
       const { currentStart, currentEnd } = rangeState;
-      console.log({ currentStart, currentEnd });
       try {
         await oo.update(currentStart, currentEnd);
         rangeState = rangeSuccessDescending({ ...rangeState, multiplier: 1 });
@@ -207,7 +206,6 @@ export const Factory = (config: Config): [ServiceFactory, Api] => {
           : defaultStartBlock;
         await queryRange(startBlock, endBlock);
         const requests = oo.listRequests();
-        console.log("requests", requests);
         const convertedRequests: SharedRequest[] = Object.values(requests).map(
           convertToSharedRequest,
         );
--- a/libs/src/oracle-sdk-v2/services/oraclev2/ethers/factory.ts
+++ b/libs/src/oracle-sdk-v2/services/oraclev2/ethers/factory.ts
@@ -197,11 +197,14 @@ export const Factory = (config: Config): [ServiceFactory, Api] => {
     } while (!rangeState.done);
   }
 
-  function queryLatestRequests(blocksAgo: number) {
+  function queryLatestRequests(blocksAgo: number, deployBlock?: number) {
     provider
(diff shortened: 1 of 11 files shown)


libs/src/oracle-sdk-v2/services/oraclev2/ethers/factory.ts
```diff
@@ -197,11 +197,14 @@ export const Factory = (config: Config): [ServiceFactory, Api] => {
...
-  function queryLatestRequests(blocksAgo: number) {
+  function queryLatestRequests(blocksAgo: number, deployBlock?: number) {
     provider
       .getBlockNumber()
       .then(async (endBlock) => {
-        const startBlock = endBlock - blocksAgo;
+        const defaultStartBlock = endBlock - blocksAgo;
+        const startBlock = deployBlock
+          ? Math.max(defaultStartBlock, deployBlock)
+          : defaultStartBlock;
```
gsteenkamp89, line 207: all of this becomes pointless after 3 days but doesn't hurt

libs/src/oracle-sdk-v2/services/oraclev3/ethers/factory.ts
```diff
@@ -217,25 +217,31 @@ export const Factory = (config: Config): [ServiceFactory, Api] => {
...
-  function queryLatestRequests(blocksAgo: number) {
+  function queryLatestRequests(blocksAgo: number, deployBlock?: number) {
     provider
       .getBlockNumber()
       .then(async (endBlock) => {
-        const startBlock = endBlock - blocksAgo;
+        const defaultStartBlock = endBlock - blocksAgo;
+        const startBlock = deployBlock
+          ? Math.max(defaultStartBlock, deployBlock)
+          : defaultStartBlock;
```
daywiss, line 244: this should be moved out of this and into caller logic imo

src/constants/env.ts
```diff
@@ -423,7 +429,7 @@ function parseEnv(env: Env): Config {
           chainId: parseInt(chainId),
           address: contractConfig.address,
           deployBlock: contractConfig.deployBlock,
-          blockHistoryLimit: 100000,
+          blockHistoryLimit: 1000000,
```
daywiss, line 432: default block limit is higher now. this may affect other things so keep an eye out if something breaks

libs/src/oracle-sdk-v2/services/oraclev3/ethers/factory.ts
```diff
@@ -217,25 +217,31 @@ export const Factory = (config: Config): [ServiceFactory, Api] => {
...
+          currentStart,
+          currentEnd,
+        );
+        updateFromEvents(eventLogs as unknown[] as OptimisticOracleEvent[]);
+        rangeState = rangeSuccessDescending({ ...rangeState, multiplier: 1 });
+      } catch (err) {
+        console.error(err, "Error loading oov ethers range");
+        rangeState = rangeFailureDescending({ ...rangeState, multiplier: 2 });
+      }
+    } while (!rangeState.done);
```
daywiss, line 235: this loop is added to progressivly load longer history. similar to the last pr update but for v3 contract

daywiss (approved):

gsteenkamp89:
@daywiss prod looks good, merging.

PR #202, opened 2024-11-12
fix: prioritize transactions found by hash

daywiss:
## motivation
we are running into an edge case, where we visit a request by tx hash, and it matches multiple txs as well as event logs. this can happen when we have a transaction requested the same time another is disputed. since we search on dispute/request/settled txs, its possible to find many matches if requests are creaetd in the same tx.

## changes
This orders any matches by a sorting preference, we prefer to locate requests if they match by request tx first, and then lower priority for dispute/settled txs.  Doing this lets us find the expected request when searching by tx and event


src/hooks/useQueryInSearchParams.ts
```diff
@@ -17,6 +17,11 @@ import { useQueries } from "./queries";
 import { useUrlBar } from "./useUrlBar";
 import { oracleEthersApiList } from "@/contexts";
 
+function isValidInteger(str: string): boolean {
+  const num = parseInt(str, 10);
+  return !isNaN(num) && num.toString() === str.trim();
+}
+
```
gsteenkamp89, line 24: can this be replaced with ?`Number.isInteger(Number(str))`

gsteenkamp89 (approved):
LGTM

src/hooks/useQueryInSearchParams.ts
daywiss: maybe so!

PR #205, opened 2024-12-23
fix: improve weights url search scorer

daywiss:
# motiviation
we are still colliding with some request urls

# changes
we scale the event log score according the the transaction match type. sorry this is a bit opaque, but it seems to solve the issue for now

### DIFF
--- a/src/hooks/useQueryInSearchParams.ts
+++ b/src/hooks/useQueryInSearchParams.ts
@@ -157,7 +157,7 @@ export function useQueryInSearchParams() {
     // adding this because sometimes event indexes are not exact
     function eventDistanceScore(value: number, target: number) {
       const distance = Math.abs(value - target);
-      if (distance === 0) return 0;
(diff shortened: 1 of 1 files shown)


src/hooks/useQueryInSearchParams.ts
```diff
@@ -157,7 +157,7 @@ export function useQueryInSearchParams() {
     // adding this because sometimes event indexes are not exact
     function eventDistanceScore(value: number, target: number) {
       const distance = Math.abs(value - target);
-      if (distance === 0) return 0;
+      if (distance === 0) return 4;
```
daywiss, line 160: we score by transaction hash, and if that matches, we score by event index. the reason why this is a distance is to handle the case were either its not supplied or off by 1 or some small margin.  here we are adding 4 to ensure a hash match and an event index match gets a higher score.

mrice32 (approved):

PR #207, opened 2025-01-15
EPIC: Multiple Values

daywiss:
# motivation
We want to add support for multiple_query identifier

# changes
This parses the requests into a standard format for display in sidebar.  

# use this branch as the EPIC for this feature, dont merge


src/helpers/converters.ts
```diff
@@ -39,6 +39,9 @@ import { erc20ABI } from "wagmi";
 import { formatBytes32String } from "./ethers";
 import { getQueryMetaData } from "./queryParsing";
 
+export const MIN_INT256 = -(BigInt(2) ** BigInt(255));
+export const MAX_INT256 = BigInt(2) ** BigInt(255) - BigInt(1);
+
```
gsteenkamp89, line 44: These constants can be imported from viem, see [here](https://github.com/wevm/viem/blob/main/src/constants/number.ts)

gsteenkamp89 (approved):

PR #208, opened 2025-01-15
feat: add mq format and parsing

daywiss:
# motivation
We need to be able to encode and decode multiple query values

# changes
This adds an encode/decode function, tested separately because could not get testing to work correctly in this repo

# tests
these pass
```ts
describe("format and parse multiple query price", () => {
  it("should format and then parse input values correctly", () => {
    const inputValues = ["10", "20", "30", "40"];
    const formattedPrice = formatMultipleQueryPrice(inputValues);
    const parsedValues = parseMultipleQueryPrice(
      formattedPrice,
      inputValues.length
    );
    expect(parsedValues).toEqual(inputValues);
  });

  it("should handle edge case of maximum input values", () => {
    const inputValues = ["0", "0", "0", "0", "0", "0", "0"];
    const formattedPrice = formatMultipleQueryPrice(inputValues);
    const parsedValues = parseMultipleQueryPrice(
      formattedPrice,
      inputValues.length
    );
    expect(parsedValues).toEqual(inputValues);
  });

  it("should handle edge case of single input value", () => {
    const inputValues = ["123456"];
    const formattedPrice = formatMultipleQueryPrice(inputValues);
    const parsedValues = parseMultipleQueryPrice(
      formattedPrice,
      inputValues.length
    );
    expect(parsedValues).toEqual(inputValues);
  });

  it("should handle edge case of input values with leading zeros", () => {
    const inputValues = ["0", "0", "0", "0", "0", "0", "1"];
    const formattedPrice = formatMultipleQueryPrice(inputValues);
    const parsedValues = parseMultipleQueryPrice(
      formattedPrice,
      inputValues.length
    );
    expect(parsedValues).toEqual(inputValues);
  });

  it("should throw an error if input values exceed 7", () => {
    const inputValues = ["1", "2", "3", "4", "5", "6", "7", "8"];
    expect(() => formatMultipleQueryPrice(inputValues)).toThrow(
      "Input values array length must not exceed 7."
    );
  });

  it("should throw an error if any value is not a valid uint32", () => {
    const inputValues = ["1", "2", "4294967296"]; // 4294967296 is 2^32, which is out of uint32 range
    expect(() => formatMultipleQueryPrice(inputValues)).toThrow(
      "Value at index 2 is not a valid uint32."
    );
  });
});
```

### DIFF
--- a/src/helpers/util.ts
+++ b/src/helpers/util.ts
@@ -298,3 +298,43 @@ export function truncateAddress(address: Address | undefined) {
     return `${address.slice(0, 5)}...${address.slice(-5)}`;
   }
 }
+
+// input user values as regular numbers
+export function formatMultipleQueryPrice(inputValues: string[]): string {
+  if (inputValues.length > 7) {
+    throw new Error("Input values array length must not exceed 7.");
+  }
+
+  let result = BigInt(0);
+
+  inputValues.forEach((value, index) => {
+    // Convert the input string value to a BigInt, interpreting it as a base-10 integer.
+    const uint32Value = BigInt(parseInt(value, 10));
+
+    // Check if the converted value is within the valid range for a uint32 (0 to 2^32 - 1).
+    if (uint32Value < 0 || uint32Value > 0xffffffff) {
+      throw new Error(`Value at index ${index} is not a valid uint32.`);
+    }
+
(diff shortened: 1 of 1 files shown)


gsteenkamp89 (approved):
awesome

PR #214, opened 2025-01-23
Multiple values fixes

gsteenkamp89:
closes UMA-2833
closes UMA-2832

### DIFF
--- a/src/hooks/actions.tsx
+++ b/src/hooks/actions.tsx
@@ -319,29 +319,6 @@ export function useProposeAction({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [proposePriceTransaction]);
 
-  if (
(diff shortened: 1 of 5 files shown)


src/hooks/actions.tsx
```diff
@@ -319,29 +319,6 @@ export function useProposeAction({
...
-
-  // TODO validate input
-  if (!isInputValid(proposePriceInput ?? "")) {
-    return {
-      title: propose,
-      disabled: true,
-      disabledReason: "Please enter a value to propose.",
-    };
-  }
-
```
gsteenkamp89, line 344: This was causing a bug in the PrimaryAction button. 
We only want to do this validation if on propose page

daywiss (approved):
great

PR #215, opened 2025-01-28
improve: allow multiple subgraph urls to be defined

daywiss:
# motivation
we want to be able to have mutliple subgraph urls to fallback to

# changes
this allows subgraphs to be defined and separated by comma in the env, and we will treat them as a priority queue of urls to try before falling back to providers


gsteenkamp89 (approved):
LGTM

PR #219, opened 2025-02-19
fix

gsteenkamp89:
## motivation

Some requests are formatted incorrectly, double quotes inside a json value are not always escaped correctly, thus we cannot parse it. 
This PR adds a utility function that will attempt to find unescaped quotes and escape them, then parse the result.


daywiss:
lets hold off on this until we know for sure we want to do this. im just afraid of introducing an anti-pattern for proposers who think sumbitting bad json is valid

gsteenkamp89:
> lets hold off on this until we know for sure we want to do this. im just afraid of introducing an anti-pattern for proposers who think sumbitting bad json is valid

Alright let's close this for now

PR #222, opened 2025-03-11
feat: add prognoze support

daywiss:
image


gsteenkamp89 (approved):
In the future we should think of separating these projects into their own config objects so we can share some of these utilities for matching a project with a proposal

PR #245, opened 2025-09-09
chore: remove custom liveness and bond entity queries

gsteenkamp89:
depends on https://github.com/UMAprotocol/subgraphs/pull/103
closes UMA-2936

## Motivation
We no longer need to query `CustomLiveness` and `CustomBond` entities separately for each request.
The `bond`, `currency` and `customLiveness` values are set directly on the `PRiceRequest` entity.
This PR removes this now obsolete logic.

### DIFF
--- a/libs/src/oracle-sdk-v2/services/managedv2/gql/queries.ts
+++ b/libs/src/oracle-sdk-v2/services/managedv2/gql/queries.ts
@@ -5,8 +5,6 @@ import type {
   OOV2GraphEntity,
   OracleType,
   PriceRequestsQuery,
-  CustomBond,
-  CustomLiveness,
 } from "@shared/types";
 import { makeQueryName } from "@shared/utils";
 import request, { gql } from "graphql-request";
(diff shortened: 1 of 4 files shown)


libs/src/oracle-sdk-v2/services/managedv2/gql/queries.ts
```diff
@@ -120,14 +120,11 @@ export async function getCustomBondForRequest(
...
+  currency: string,
+): Promise<Pick<CustomBond, "customBond" | "currency"> | undefined> {
   const query = gql`
     query GetCustomBond {
-      customBonds(where: { requester: "${requester}", identifier: "${identifier}", ancillaryData: "${ancillaryData}" }) {
-        id
-        requester
-        identifier
-        ancillaryData
+      customBonds(where: { requester: "${requester}", identifier: "${identifier}", ancillaryData: "${ancillaryData}", currency: "${currency}" }) {
```
gsteenkamp89, line 127: must match the `currency` of the request

libs/src/oracle-sdk-v2/services/managedv2/gql/queries.ts
```diff
@@ -141,22 +138,18 @@ export async function getCustomBondForRequest(
...
-) {
+): Promise<Pick<CustomLiveness, "ancillaryData"> | undefined> {
   const query = gql`
     query GetCustomLiveness {
       customLivenesses(where: { requester: "${requester}", identifier: "${identifier}", ancillaryData: "${ancillaryData}" }) {
-        id
-        requester
-        identifier
-        ancillaryData
         customLiveness
```
gsteenkamp89, line 153: we only care about liveness

PR #278, opened 2026-05-22
Port Polymarket bulletin owner remap

pumpedlunch:
## Summary
- Port Polymarket bulletin owner remap behavior from voter-dapp PR #455
- Query both current and deprecated bulletin owners for the deprecated initializer
- Filter deprecated-owner updates after cutoff timestamp 1779436800, then merge and sort updates

## Verification
- yarn test src/projects/polymarketBulletinOwners.test.ts
- yarn prettier --check src/projects/polymarketBulletinOwners.ts src/projects/polymarketBulletinOwners.test.ts src/components/Panel/PolymarketTextData.tsx --config ./prettier.config.mjs
- yarn lint
- yarn tsc --noEmit --pretty false


src/projects/polymarketBulletinOwners.ts
```diff
@@ -0,0 +1,85 @@
...
+  },
+};
+
+function getTimestampNumber(
+  timestamp: RawPolymarketBulletinUpdate["timestamp"],
+) {
+  return Number(timestamp.toString());
+}
+
+export function resolveBulletinOwner(initializer: Address): Address {
```
droplet-rl, line 39: `resolveBulletinOwner` is exported but only consumed by the test file. If there's no near-term caller, consider dropping the export (or the function) to avoid keeping a public surface that has to be maintained. Non-blocking — fine to keep if the source voter-dapp PR has external consumers you're planning to port next.

src/projects/polymarketBulletinOwners.ts
```diff
@@ -0,0 +1,85 @@
...
+  ownerQueries: readonly BulletinOwnerQuery[],
+): TUpdate[] {
+  return updatesByOwner
+    .flatMap((updates, ownerIndex) => {
+      const ownerQuery = ownerQueries[ownerIndex];
+
+      return updates.filter((update) => {
+        return (
+          ownerQuery?.maxTimestamp === undefined ||
+          getTimestampNumber(update.timestamp) <= ownerQuery.maxTimestamp
```
droplet-rl, line 77: Minor: `getTimestampNumber(update.timestamp)` is recomputed once in the filter and twice per comparison in the sort. For these small lists it's irrelevant, but you could `.map` to a pre-numericized shape once before sorting to avoid the repeated `Number(toString())` and keep the sort comparator tight. Optional.

src/projects/polymarketBulletinOwners.test.ts
```diff
@@ -0,0 +1,105 @@
...
+        "current-old",
+      ),
+      update(
+        deprecatedPolymarketBulletinOwnerCutoffTimestamp - 10,
+        "deprecated-old",
+      ),
+      update(deprecatedPolymarketBulletinOwnerCutoffTimestamp + 20, "current"),
+    ]);
+  });
+});
```
droplet-rl, line 105: Worth a one-line test that calling `mergePolymarketBulletinUpdates` with the single-query result from an unmapped initializer passes updates through unfiltered — it's the common path in production and currently only the deprecated-initializer branch is exercised end-to-end through merge. Non-blocking.

src/components/Panel/PolymarketTextData.tsx
```diff
@@ -30,15 +35,26 @@ export function useGetUpdates(props: {
...
+    allowFailure: false,
+    contracts: ownerQueries.map((ownerQuery) => ({
+      address,
+      abi: polymarketBulletinAbi,
+      functionName: "getUpdates",
+      // bulletin contract on polygon
+      chainId: 137,
+      args: [questionId as Address, ownerQuery.owner],
+    })),
     enabled: !!address && !!questionId && !!ownerAddress,
```
droplet-rl, line 52: When `ownerAddress` is undefined, `ownerQueries` is `[]` so `contracts` is `[]` and `enabled` is false — wagmi will short-circuit and the `select` callback isn't invoked, so this is safe. Just flagging that the empty-array + disabled combo is the intended idle state, in case anyone trips over it later.

droplet-rl (approved):
Approving. Clean port of the voter-dapp bulletin owner remap.

**What's right**
- `useContractRead` → `useContractReads({ allowFailure: false })` lets one hook fan out to both the current owner and the deprecated initializer. With `allowFailure: false` the `select` callback can treat the result as a flat `RawPolymarketBulletinUpdate[][]`.
- `getBulletinOwnerQueries` orders queries `[current, deprecated]` and the merge filter keys off the matching index — the cutoff (`maxTimestamp`) is only applied to the deprecated owner, exactly as described.
- `resolveBulletinOwner` / record lookup is case-insensitive via `toLowerCase()`, and the deprecated-initializer record key is stored lowercased.
- The React `key` switch to `${timestamp}-${index}` correctly handles same-timestamp collisions that can now happen since updates from two owners are merged.
- Tests cover the mapped/unmapped resolve paths, the query-shape branching, and the filter+merge+sort behavior.

**Notes / nits (non-blocking)**
- `polymarket.ts` shows a stray removal of `0xac9930b2…825294` in the 2-dot diff, but the 3-dot (merge-base) diff is clean — it's just branch lag. Make sure `update-integrations` is brought up to date with master before merging so that initializer isn't accidentally dropped.
- See inline comments for a couple of small optional follow-ups.

mrice32 (commented):
LGTM!

